PR #55, opened 2014-01-22
Added support for Account

dylandrop:
Adds `Stripe::Account.retrieve`. Note:  `Stripe::Account.retrieve(access_token)` also needs to be implemented (as is in the Stripe Connect flow), but I wasn't sure how to do this. The problem is that although it is possible to call this from the `stripe-ruby` gem, it isn't covered in the Stripe docs and I'm having trouble finding what the route is for this. Furthermore, I'm not sure how I would implement this since it's not as simple as following the formula of the last few API endpoints (or so it seems). Perhaps someone could help out here.


alexmamonchik:
@dylandrop @woodhull your commits and commits from #218 merged to `accounts` branch. I going to fix that branch and close this PR and #218. If you want to improve something - welcome to `accounts`. Later it will be merged to `master`

alexmamonchik:
fixed by #241

Issue #72, opened 2014-03-08
Stripe::Customer.create with a coupon does not return a discount


andyjbas:
Currently when a `Stripe::Customer` is created with a `coupon` attribute, a discount is not created and returned with the customer response. Instead, a `coupon` property is returned, but this does not match Stripe's live response.

Example:

```ruby
coupon = 'stripe_4_lyfe'

Stripe::Coupon.create(
  percent_off: 25,
  duration: 'repeating',
  duration_in_months: 3,
  id: coupon
)

customer_card = {
  number: 4242424242424242,
  exp_month: 2,
  exp_year: Time.now.year.next,
  cvc: "314"
}

Stripe::Customer.create(card: customer_card, coupon: coupon)
```

With the mock server running:

```json
{
  "id": "test_cus_5",
  "email": "stripe_mock@example.com",
  "description": "an auto-generated stripe customer data mock",
  "object": "customer",
  "created": 1372126710,
  "livemode": false,
  "delinquent": false,
  "discount": null,
  "account_balance": 0,
  "cards": {"object":"list","count":1,"url":"/v1/customers/test_cus_5/cards","data":[{"id":"test_cc_6","object":"card","last4":"4242","type":"Visa","exp_month":4,"exp_year":2016,"fingerprint":"wXWJT135mEK107G8","customer":"test_cus_5","country":"US","name":"Johnny App","address_line1":null,"address_line2":null,"address_city":null,"address_state":null,"address_zip":null,"address_country":null,"cvc_check":null,"address_line1_check":null,"address_zip_check":null}]},
  "default_card": "test_cc_6",
  "coupon": "stripe_4_lyfe",
  "subscriptions": {"data":[],"count":0,"object":"list","url":"/v1/customers/test_cus_5/subscriptions"}
}
```

Without the mock server running:

```json
{
  "id": "cus_3d82VayotGv2d6",
  "object": "customer",
  "created": 1394313303,
  "livemode": false,
  "description": null,
  "email": null,
  "delinquent": false,
  "metadata": {},
  "subscriptions": {"object":"list","count":0,"url":"/v1/customers/cus_3d82VayotGv2d6/subscriptions","data":[]},
[...]

PR #74, opened 2014-03-14
Create customer with coupon discount

guilleart:
Referencing https://github.com/rebelidealist/stripe-ruby-mock/issues/72

Customers can be created now passing a coupon id. Created customers will have the discount field set to the coupon as long as the coupon passed existed in the first place.

Any comments let me know.


hcarver:
I added the ability to create subscriptions with coupons in [https://github.com/hcarver/stripe-ruby-mock](my fork). But the coupons aren't actually applied anywhere yet - just accepted and returned.

alexmamonchik:
looks like most of initial code already implemented by @jrmyward

PR #117, opened 2014-08-06
Update Branding and Funding Fields to match current Stripe API

jcope2013:
previously the branding field was instead under "type" and the funding field didn't exist, these have since been added to the Stripe API so I added fields for them


EddM:
Thanks for your help, but this has now been covered by some other PRs

PR #118, opened 2014-08-11
Remove spec warnings

pjambet:
Remove some RSpec 3 warnings

Convert `be_false` to `eq(false)`
Convert `be_true` to `eq(true)`


alexmamonchik:
fixed by another PR

PR #119, opened 2014-08-12
Add refunds

pjambet:
This pull request adds support for refunds, as of API version 2014-06-17[1] 

In order to keep backward compatibility I added a version to api handler.
The idea is that all new backward incompatible handlers, such as
`klass.add_handler 'post /v1/charges/(.*)/refunds', :create_refund, '2014-06-17'`
should have a version defined.

The version is either defined through `--api-version` when running the server through the binstub or if you're not running the server, simply by doing `StripeMock.version = '2014-06-14'`.

The pull request is not 100% ready yet, as you can see the Refund handler class is not finished yet. So let me know if you're ok with my implementation, and if so, I'll wrap up everything so we can merge this.

Again, thanks for this project, it is really helpful for us!

PS: I realize that I should add specs for the `handler_for_method_url` method. I'll try to do that as soon as I can.

[1] https://stripe.com/docs/upgrades#2014-06-17

### DIFF
--- a/bin/stripe-mock-server
+++ b/bin/stripe-mock-server
@@ -4,6 +4,8 @@ lib = File.expand_path(File.dirname(__FILE__) + '/../lib')
 $LOAD_PATH.unshift(lib) if File.directory?(lib) && !$LOAD_PATH.include?(lib)
 
 require 'trollop'
+require 'stripe_mock'
+require 'stripe_mock/server'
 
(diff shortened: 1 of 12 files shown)


arthurnn:
Is this still workable? should we close? should we re-work it?

pjambet:
I definitely won't have the time to re-work this. I'm going to close this for now.

PR #132, opened 2014-09-03
type of credit card is actually brand according to API

ilyakatz:
According to the [docs](https://stripe.com/docs/api/ruby#retrieve_card), credit card type is called "brand"

Alt text


alexmamonchik:
already implemented

PR #141, opened 2014-10-09
Add limit param to list_invoices

antoinelyset:
I think this is enough to implement the documented behavior : https://stripe.com/docs/api/curl#list_customer_invoices


EddM:
Fixed with #172

antoinelyset:
Thanks !

PR #143, opened 2014-10-22
Update the mocked card objects to use brand instead of type.

jbranchaud:
The Stripe API renamed Card `type` to Card `brand` ([see discussion here](https://groups.google.com/a/lists.stripe.com/forum/#!topic/api-discuss/4VULNor--Bk)). This PR updates the mocked Card objects accordingly.


EddM:
Thanks, looks like this was fixed in 8913129 though

PR #151, opened 2014-11-12
Return List objects when scoped to customer

nikolai-b:
When a customer does not exist this will now raise the correct error.  If the customer does exist it will return a list object.

### DIFF
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -100,7 +100,7 @@ module StripeMock
 
     def self.mock_charge_array
       {
-        :data => [test_charge, test_charge, test_charge],
+        :data => [],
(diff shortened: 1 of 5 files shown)


EddM:
Awesome, thanks! Will there ever be a case where the Stripe API *shouldn't* return a `list`, though?

nikolai-b:
Do you mean that `Stripe::Charge.all` should always return a `Stripe::ListObject`?  That's a good point.  Do you want me to add that to the PR?

EddM:
That would be great if you could

EddM:
Fixed in #172

nikolai-b:
sorry didn't get time.  Seems like you had a more comprehensive solution.

Issue #157, opened 2014-12-21
Creating tokens with create_test_helper doesn't accept last4 params, but does without create_test_helper


spencerldixon:
I'm wanting to create a token with specific last4 digits to assert they get updated when a user updates their card. Doing the following, will always generate a token with the last4 of '4242', regardless of what I pass in...

```
StripeMock.start
stripe_helper = StripeMock.create_test_helper
stripe_helper.create_plan(id: 'basic')
@token = stripe_helper.generate_card_token(last4: "9191", exp_year: 1984)
```
However, ignoring the create_test_helper and using the old method like below, will create the token with the custom last4 successfully...

`@token = stripe_helper.generate_card_token(last4: "9191", exp_year: 1984)`

Am I doing something wrong or is this a legitimate bug and is there any difference in using create_test_helper for tokens rather than the direct method?

PR #159, opened 2014-12-22
add card merge for dealing with number and last4

nikolai-b:
adding a Util to deal with merging in card numbers and last4, see #157


mindeavor:
Well done, thank you.

PR #162, opened 2015-01-05
allow creation of invoice line items from invoice

ryanong:
This will allow users to use `invoice.lines.create`


ryanong:
never mind, interaction is a bit too complex to safely mock.

PR #164, opened 2015-01-12
Add `tax_percent` to Data.mock_subscription

jbourassa:
The field is listed in the API documentation but not in the mocks.

PS: Good job on stripe-ruby-mock, this is incredibly helpful and easy to use.


mindeavor:
Thank you!

Issue #128, opened 2014-08-26
subscriptions "count" should be changed to "total_count"


ygamretuta:
This is the JSON response of stripe-mock when I fetch subscriptions for a customer

```
#<Stripe::ListObject:0x85c7038c> JSON: {
  "object": "list",
  "count": 1,
  "url": "/v1/customers/test_cus_3/subscriptions",
  "data": [
    {"id":"test_su_4","current_period_start":1409089581,"current_period_end":1411767981,"status":"active","plan":{"id":"basic","interval":"month","name":"Basic","amount":500,"currency":"usd","object":"plan","livemode":false,"interval_count":1,"trial_period_days":null},"cancel_at_period_end":false,"canceled_at":null,"ended_at":null,"start":1308595038,"object":"subscription","trial_start":null,"trial_end":null,"customer":"test_cus_3","quantity":1}
  ]
}
```

this is the official response according to docs
```
"subscriptions": {
    "object": "list",
    "total_count": 1,
    "has_more": false,
    "url": "/v1/customers/cus_4dJQFvyQzAVlk7/subscriptions",
    "data": []
```

EddM:
Fixed by @michald in #166

PR #166, opened 2015-01-22
Invalid customer/cards count attribute

michald:
Hi there,

I wasn't be able to write my tests properly without this fix. It's related to `count` attribute in `customer/cards` object which should be replaced by `total_count`. Also I wasn't be able to increase number of cards because of this wrong attribute.

https://stripe.com/docs/api#retrieve_customer

### DIFF
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -16,7 +16,7 @@ module StripeMock
         account_balance: 0,
         cards: {
           object: "list",
-          count: cards.count,
+          total_count: cards.size,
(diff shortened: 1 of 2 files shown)


theoretick:
+1.  According to the README target API version is `STRIPE API TARGET VERSION: 2014-06-17`, which uses `total_count`, not `count`.  I looked into it and it seems that `count` was dropped in `v2014-03-28`.

This fix addresses https://github.com/rebelidealist/stripe-ruby-mock/issues/128

EddM:
Thanks @michald. Although this is a potentially breaking change, it is the expected behaviour for the API version currently being targeted.

Issue #161, opened 2015-01-01
Stripe::Customer.create does not return Subscription data


AMPearlman:
When calling Stripe::Customer.create(data) against the Stripe API, newly created "subscription" data is returned. (Not to be confused with "subscriptions" which contains all of a user's subscriptions.)

Example:
#<Stripe::Customer:0x3fe55cfe6aec id=cus_5QpquIeafKVluu> JSON: {
  "id": "cus_5QpquIeafKVluu",
  "object": "customer",
  "created": 1420077299,
  "livemode": false,
  "description": "test.com",
  "email": "test@test.com",
  "delinquent": false,
  "metadata": {},
  "subscriptions": {"object":"list","total_count":1,"has_more":false,"url":"/v1/customers/cus_5QpquIeafKVluu/subscriptions","data":[{"id":"sub_5QpqlE9O49vUN7","plan":{"id":"single","interval":"year","name":"Single","created":1400439973,"amount":9999,"currency":"usd","object":"plan","livemode":false,"interval_count":1,"trial_period_days":null,"metadata":{},"statement_descriptor":null,"statement_description":null},"object":"subscription","start":1420077299,"status":"active","customer":"cus_5QpquIeafKVluu","cancel_at_period_end":false,"current_period_start":1420077299,"current_period_end":1451613299,"ended_at":null,"trial_start":null,"trial_end":null,"canceled_at":null,"quantity":1,"application_fee_percent":null,"discount":null,"tax_percent":null,"metadata":{}}],"count":1},
  "discount": null,
  "account_balance": 0,
  "currency": "usd",
  "cards": {"object":"list","total_count":1,"has_more":false,"url":"/v1/customers/cus_5QpquIeafKVluu/cards","data":[{"id":"card_5QpqDwZkqzg2HG","object":"card","last4":"4242","brand":"Visa","funding":"credit","exp_month":1,"exp_year":2016,"fingerprint":"ptgEEAD7q8eXdxdc","country":"US","name":null,"address_line1":null,"address_line2":null,"address_city":null,"address_state":null,"address_zip":null,"address_country":null,"cvc_check":"pass","address_line1_check":null,"address_zip_check":null,"dynamic_last4":null,"customer":"cus_5QpquIeafKVluu","type":"Visa"}],"count":1},
  "default_card": "card_5QpqDwZkqzg2HG",
[...]

Issue #160, opened 2014-12-23
Customers.invoice returns array, not Stripe::ListObject


kareem:
When I get a customer's invoices from the Stripe API, I get this:

`irb(main):009:0> c.invoices`

`=> #<Stripe::ListObject:0x3fc89d49b1a0> JSON: {
  "object": "list",
  "has_more": true,
  "url": "/v1/invoices",
  "data": [ Array_of_Stripe::Invoice_objects]
}`

When I get a customer's invoices from this mocking gem, I get:

`(rdb:1) pp @stripe_customer.invoices`

`[#<Stripe::Invoice:0x3fd8455a50ac id=test_in_3> JSON: {
  "id": "test_in_3",
  "date": 1349738950,
  "period_end": 1349738950,
  "period_start": 1349738950,
  "lines": {"object":"list","count":1,"url":"/v1/invoices/test_in_3/lines","data":[{"id":"ii_test","object":"line_item","type":"invoiceitem","livemode":false,"amount":1000,"currency":"usd","proration":false,"period":{"start":1349738920,"end":1349738920},"quantity":null,"plan":null,"description":"Test invoice item","metadata":{}}]},
  "subtotal": 1000,
  "total": 1000,
  "customer": "stripe_customer_id_1",
  "object": "invoice",
  "attempted": false,
  "closed": false,
  "paid": false,
  "livemode": false,
  "attempt_count": 0,
  "amount_due": 1000,
  "currency": "usd",
  "starting_balance": 0,
  "ending_balance": null,
  "next_payment_attempt": 1349825350,
  "charge": null,
  "discount": null,
  "subscription": null
}`

Note the difference; the API returns a `Stripe::ListObject`, while the mocking gem returns an array of `Stripe::Invoices`.

The problem is that my code assumes the `StripeInvoices` are in the `Stripe::ListObject.data` array, so I can't write tests against this part of my code.

EddM:
Just FYI to those who read this, PR #151 should address this

EddM:
Fixed in #172

PR #172, opened 2015-01-28
List objects and pagination

EddM:
There's a bunch of responses that StripeMock returns which are simply arrays, not lists like the live Stripe API returns. There are a some issues opened about this, including some PRs which patch only parts of this (e.g. #161, #151, #160). This provides a common `List` class which can wrap any array of mock Stripe objects into a list and additionally add pagination functionality

### DIFF
--- a/lib/stripe_mock.rb
+++ b/lib/stripe_mock.rb
@@ -5,10 +5,11 @@ require 'dante'
 require 'stripe'
 
 require 'stripe_mock/version'
-require 'stripe_mock/data'
-require 'stripe_mock/list'
 require 'stripe_mock/util'
 require 'stripe_mock/error_queue'
+
+require 'stripe_mock/data'
+require 'stripe_mock/data/list'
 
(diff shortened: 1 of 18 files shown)


mindeavor:
Great work! A couple minor things (I can't comment on a line on mobile, apparently):

Does this charge id need to be hardcoded in?

```ruby
list = StripeMock::Data::List.new(data, starting_after: "test_ch_261")
```

And there's one test desc that needs to be updated (count -> limit):

```ruby
context "when passing count" do
```

Otherwise I'd say it's clear to merge. Out of curiosity, did you base the new List class off of Stripe's ruby gem's implementation?

EddM:
> Does this charge id need to be hardcoded in?

Yeah good point, if the id generation changed in the future, this would fail unnecessarily

> did you base the new List class off of Stripe's ruby gem's implementation?

Not really, more just based off the general pattern the live API returns -- and the existing `list`s already present in some of StipeMock's mock responses

PR #173, opened 2015-02-02
removed request for additional core contributor

mlpinit:
- I was interested in helping out with the project as well but then I realized
you already found a colaboraor. If that's correct here's a pr to remove the
request from the README page.


mindeavor:
You are correct, thank you!

PR #177, opened 2015-02-03
Allow tests to retrieve webhook payloads and not just Stripe::Events

nhocki:
With this change, people can easily get the webhook payload to test their handlers without having an Stripe::Event object.

It's basically extracting reading the json files to its own method instead of having them in the `mock_webhook_event`.


sebasoga:
+1

mindeavor:
Looks great, thank you!

PR #179, opened 2015-02-11
Allow to retrieve all cards from recipients

nhocki:
With this PR we can now get the cards from a recipient similar to what is possible for customers.


sebasoga:
:+1:

EddM:
Great, thanks

PR #180, opened 2015-02-12
Subscriptions attribute count to total_count

passbe:
The major stripe upgrade of 2014-03-28 replaced the lists attribute with an optional total_count. Updated subscriptions to reflect this change.

https://stripe.com/docs/upgrades#2014-03-28


EddM:
Thanks! AFAIK this is the last instance we missed where it was returning `:count` instead of `:total_count`. Hopefully responses will include this dynamically through `StripeMock::Data::List` soon

PR #181, opened 2015-02-15
Account for new param in Stripe.request

hayesr:
Looks like Stripe added a param to their request method ( [see here](https://github.com/stripe/stripe-ruby/blob/ef249b7cebf9da298ed0ba7c43555e93c0295d6d/lib/stripe.rb#L70) )

When calling Stripe::Customer.create with StripeMock started, an extra param is passed causing an ArgumentError and general pain and sadness. ( [see this line](https://github.com/stripe/stripe-ruby/blob/4d611c62f71149d5340ecee6b2c35345cb35b84d/lib/stripe/api_operations/request.rb#L15) )

So this small change accounts for that. I didn't add any tests since no functionality changed. Current tests pass.

### DIFF
--- a/lib/stripe_mock/instance.rb
+++ b/lib/stripe_mock/instance.rb
@@ -58,7 +58,7 @@ module StripeMock
       @base_strategy = TestStrategies::Base.new
     end
 
-    def mock_request(method, url, api_key, params={}, headers={})
+    def mock_request(method, url, api_key, params={}, headers={}, api_base_url=nil)
       return {} if method == :xtest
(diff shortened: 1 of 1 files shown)


EddM:
Feels weird for this param to be useless and just there for compatibility's sake, but I guess StripeMock genuinely doesn't care about it. Thanks!

hayesr:
It is a bit weird. An alternative could be to collect all the params in a splat. This would be future-proof, but not really cleaner.

PR #187, opened 2015-02-17
Allow the tax_percent to be set when the subscription is created.

gravitystorm:
Although the tax_percent has a default value, the custom_subscription_params method wasn't allowing it to be set when creating a subscription with a non-default tax_rate.


mindeavor:
Thank you!

Issue #71, opened 2014-03-05
Handle invalid amount on Stripe::Charge calls


larslevie:
The Stripe API will reject attempted charges with a negative, non-integer, or zero amount. Currently the mock treats these charges as valid.

PR #190, opened 2015-02-24
Disallow Stripe::Charge without required params or non positive-integer amounts

theoretick:
This addresses #71, stripe charges should not allow non-positive-integer amounts:
```
[7] pry »  Stripe::Charge.create(amount: -1, currency: 'usd', customer: stripe_customer_id)
Stripe::InvalidRequestError: Invalid positive integer
from /home/theoretick/.gem/ruby/2.1.2/gems/stripe-1.16.0/lib/stripe.rb:232:in `handle_api_error'
```

Breaking change: I had to change quite a few references with this one since `Stripe::Charge.create` without params was used abundantly elsewhere in the test suite.

Unfortunately, it kills the convenience of easy charge inits but this accurately reflects stripe API usage.

I live-tested these (leading to second commit to clarify slightly different error message), but still had a few unrelated failures i couldn't resolve that did not go away when i reverted changes:

```ruby
$ bundle exec rspec -t live
Running **live** tests against Stripe...
Run options:
  include {:live=>true}
  exclude {:mock_server=>true, :oauth=>true}

StripeMock::Instance
  behaves like Refund API
    creates a stripe refund with the charge ID
    creates a refund off a charge (FAILED - 1)
    handles multiple refunds (FAILED - 2)

Failures:

  1) StripeMock::Instance behaves like Refund API creates a refund off a charge
     Failure/Error: refund = charge.refunds.create(amount: 555)
     NoMethodError:
       undefined method `create' for []:Array
     Shared Example Group: "Refund API" called from ./spec/support/stripe_examples.rb:19
     # ./spec/shared_stripe_examples/refund_examples.rb:63:in `block (2 levels) in <top (required)>'

  2) StripeMock::Instance behaves like Refund API handles multiple refunds
     Failure/Error: refund_1 = charge.refunds.create(amount: 300)
     NoMethodError:
       undefined method `create' for []:Array
     Shared Example Group: "Refund API" called from ./spec/support/stripe_examples.rb:19
     # ./spec/shared_stripe_examples/refund_examples.rb:73:in `block (2 levels) in <top (required)>'

Finished in 6.63 seconds (files took 0.39599 seconds to load)
3 examples, 2 failures

Failed examples:

rspec ./spec/shared_stripe_examples/refund_examples.rb:58 # StripeMock::Instance behaves like Refund API creates a refund off a charge
rspec ./spec/shared_stripe_examples/refund_examples.rb:68 # StripeMock::Instance behaves like Refund API handles multiple refunds
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ it "mocks a declined card error" do
   # Prepares an error for the next create charge request
   StripeMock.prepare_card_error(:card_declined)
 
-  expect { Stripe::Charge.create(amount: 1) }.to raise_error {|e|
+  expect { Stripe::Charge.create(amount: 1, currency: 'usd') }.to raise_error {|e|
     expect(e).to be_a Stripe::CardError
     expect(e.http_status).to eq(402)
     expect(e.code).to eq('card_declined')
@@ -171,7 +171,7 @@ it "raises a custom error for specific actions" do
 
   StripeMock.prepare_error(custom_error, :new_customer)
 
-  expect { Stripe::Charge.create(amount: 1) }.to_not raise_error
+  expect { Stripe::Charge.create(amount: 1, currency: 'usd') }.to_not raise_error
   expect { Stripe::Customer.create }.to raise_error {|e|
     expect(e).to be_a StandardError
     expect(e.message).to eq("Please knock first.")
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ it "mocks a declined card error" do
   # Prepares an error for the next create charge request
   StripeMock.prepare_card_error(:card_declined)
 
-  expect { Stripe::Charge.create }.to raise_error {|e|
+  expect { Stripe::Charge.create(amount: 1) }.to raise_error {|e|
     expect(e).to be_a Stripe::CardError
     expect(e.http_status).to eq(402)
     expect(e.code).to eq('card_declined')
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by EddM** [comment]
Why did the tests that created param-less `Stripe::Charge`s fail after your changes? I can't see why they would.

They probably *should* fail, since the Stripe API requires an `amount` (and `currency`) for charges, but StripeMock doesn't currently enforce that

**Comment by theoretick** [comment]
Haha, sheesh, I guess I updated the calls but left off that actual presence check.

Well since this fix doesn't _need_ to introduce that breaking change I guess I could either (A) revert the param-less `Stripe::Charge` calls and leave them as they are, or (B) I could add an explicit failure for "no amount provided" to mirror Stripe API (and possibly `currency`).

What do you think?  I suppose (B) falls outside the scope of this ticket, but maybe its worth doing while I have it half done.

**Comment by EddM** [comment]
Ah, I see - yeah, checking for the presence of those required fields sounds like a good idea to me, if you don't mind

**Comment by theoretick** [comment]
K, added requirements for `amount` and `currency` params.  I was going to add minimum amount as well:
```ruby
elsif amount_below_minimum_charge?(params)
  raise Stripe::InvalidRequestError.new('Amount must be at least 50 cents', 'amount', 400)
```
but it gets a bit complex with currency conversions and I'm unsure how much internationalization is desired/enabled in stripe-ruby-mock currently.  I'll make a new ticket for adding it in the future.

mindeavor:
Looks great!

theoretick:
:tada: :tada:

PR #191, opened 2015-02-27
Stripe API update [Breaking change]

### LINKED ISSUE #189: Customer needs to have source instead of card now
Stripe has made a pretty major change to how customers cards and/or other payment sources are set and retrieved. Charge and default_charge have been swapped with source and default_source, respectively. Which, if you allow more payment types than just cards, returns a list of different payment types rather than one. See https://stripe.com/docs/upgrades

nhocki:
This PR will update the Stripe API to the most recent version ([changes here](https://stripe.com/docs/upgrades?since=2015-01-26#whats-changed-since-2015-01-26)).

* This is a breaking change. `Customer`s and `Charge`s will now have `sources` instead of cards.
* This still assumes that all `sources` of a customer are cards, there's no Bitcoin support right now.
* This still has the `customers/cards` endpoints since the Stripe gem has that problem. See stripe/stripe-ruby#215

* No version bump here, I'll leave that to the maintainers.

This fixes #189

### DIFF
--- a/lib/stripe_mock.rb
+++ b/lib/stripe_mock.rb
@@ -43,6 +43,7 @@ require 'stripe_mock/request_handlers/validators/param_validators.rb'
 
 require 'stripe_mock/request_handlers/charges.rb'
 require 'stripe_mock/request_handlers/cards.rb'
+require 'stripe_mock/request_handlers/sources.rb'
 require 'stripe_mock/request_handlers/customers.rb'
 require 'stripe_mock/request_handlers/coupons.rb'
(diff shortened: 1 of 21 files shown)


mindeavor:
Thanks for the PR! Can you write comments around the code that will become irrelevant once Stripe resolves stripe/stripe-ruby#215?

nhocki:
@mindeavor so Stripe just released a new version `1.20.1` that fixes the problem. I've removed the code and bumped the required Stripe version.

mindeavor:
Thank you!

plicjo:
Good work!

PR #200, opened 2015-03-11
fixing for stripe v 1.20.2

slavad:
With Stripe v 1.20.2  current version of mock does not retrieve plan name correctly.
E.g. if ``params[:plan] == {}`` and `` subscription[:plan][:id] == 'some-plan'``, the result will be ``{}``, but not ``some-plan``

### DIFF
--- a/lib/stripe_mock/request_handlers/subscriptions.rb
+++ b/lib/stripe_mock/request_handlers/subscriptions.rb
@@ -61,7 +61,11 @@ module StripeMock
         end
 
         # expand the plan for addition to the customer object
-        plan_name = params[:plan] || subscription[:plan][:id]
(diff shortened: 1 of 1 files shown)


mindeavor:
This seems to break the build: https://travis-ci.org/rebelidealist/stripe-ruby-mock/jobs/53978624

peterkeen:
FYI this is biting Payola's tests pretty hard: https://circleci.com/gh/peterkeen/payola/280

Forcing stripe-ruby-mock v2.1.0 to use 1.20.2 results in 19 test failures, all by itself. Master has a whole mess of failures as well.

slavad:
I see. Looks like it needs more fixing.

PR #218, opened 2015-04-24
Add Accounts

sax:
This pull request adds the ability to mock Stripe::Account records (https://stripe.com/docs/api#account).

Included is a change to recursively merge hashes in other data mocks. We were not sure if it was an oversight or not that these were not recursively merged, but had a lot of issues where overriding specific nested parameters would cause method missing errors on the resulting objects.

### DIFF
--- a/lib/stripe_mock.rb
+++ b/lib/stripe_mock.rb
@@ -41,6 +41,7 @@ require 'stripe_mock/request_handlers/helpers/token_helpers.rb'
 
 require 'stripe_mock/request_handlers/validators/param_validators.rb'
 
+require 'stripe_mock/request_handlers/accounts.rb'
 require 'stripe_mock/request_handlers/charges.rb'
 require 'stripe_mock/request_handlers/cards.rb'
 require 'stripe_mock/request_handlers/sources.rb'
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -85,8 +85,8 @@ module StripeMock
           avs_failure: false
         },
         keys: {
-          secret: nil,
-          publishable: nil
+          secret: 'SECRETKEY',
+          publishable: 'PUBLISHABLEKEY'
         }
       }.merge(params)
     end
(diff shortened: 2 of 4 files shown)


sax:
We saw that there was a similar feature in #55, but the Stripe gem and APIs have changed since then. Account used to be a singleton object representing the current user's account. Now there is an API for managing connected accounts in a marketplace context..

alexmamonchik:
@sax thanks for PR. Can you review `accounts` branch. I took and merged your commits and commits from #55

sax:
@alexmamonchik sorry it's taken me so long to get back to you!

I'm not sure that the merged pull requests would actually work in our code base. It looks like #55 is using the Account in terms of the singular Account record of the current API user. In the newer Stripe code, Account is used in two ways:

* without an ID, in reference to the current account
* with an ID, in reference to a managed account in a marketplace context

We're using the latter bit of code. We run a marketplace, where we create Stripe Account records for the vendors that we take and manage payments for. There's a lot more fields involved, and the nesting of the fields makes it kind of a pain to work without deep merging the data passed into `#mock_account`.

alexmamonchik:
@sax can you show few examples how you use accounts?

jonbcampos:
+1 would like to see this added, but would also like to see "external_accounts" added to the account object

    ...
    external_accounts:{
        object: "list",
        data: [],
        has_more: false,
        total_count: 0,
        url: "/v1/accounts/#{id}/external_accounts"
    }
    ...

into the mock_account method

alexmamonchik:
@sax merged partially. Please, take a look and confirm that we can close this PR.

alexanmtz:
I've checked this branch and it works well for me in my managed account tests. I'm just missing something like `StripeMock.prepare_account_error(:invalid_request_error)`, but I could handle with custom errors. I'm making the error handler as for card and then I can create a pull request for this after this code is merged.

csalvato:
Closing as stale.

PR #225, opened 2015-05-10
refactoring live tests for coupons

alexmamonchik:
Fix commet from @mindeavor

### DIFF
--- a/lib/stripe_mock/test_strategies/base.rb
+++ b/lib/stripe_mock/test_strategies/base.rb
@@ -21,6 +21,37 @@ module StripeMock
         stripe_token.id
       end
 
+      def create_coupon_params(params = {})
+        {
+          id: '10BUCKS',
+          amount_off: 1000,
+          currency: 'usd',
+          max_redemptions: 100,
+          metadata: {
+            created_by: 'admin_acct_1'
+          },
+          duration: 'once'
+        }.merge(params)
+      end
+
+      def create_coupon_percent_of_params(params = {})
+        {
+          id: '25PERCENT',
+          percent_off: 25,
+          redeem_by: nil,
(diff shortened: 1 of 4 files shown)


lib/stripe_mock/test_strategies/base.rb
```diff
@@ -21,6 +21,37 @@ def generate_card_token(card_params={})
...
+          currency: 'usd',
+          max_redemptions: 100,
+          metadata: {
+            created_by: 'admin_acct_1'
+          },
+          duration: 'once'
+        }.merge(params)
+      end
+
+      def create_coupon_percent_of_params(params = {})
```
mindeavor: Where does this get used?

alexmamonchik: [here] (https://github.com/rebelidealist/stripe-ruby-mock/blob/master/lib/stripe_mock/test_strategies/live.rb#L21). But that method
```ruby
def create_coupon(params={})
  delete_coupon create_coupon_params(params)[:id]
  super
end
```
allows create coupon with `amount_off` only. I have method `create_coupon_percent_of_params` and I going to use it in next commits. Also I found strange behavior when add `percent_off` coupon to subscription. That is not confirmed bug yet.

mindeavor: Sorry, I wasn't specific. I do see that `create_coupon_params` gets used, but I couldn't find where `create_coupon_percent_of_params` got used / tested. Maybe just forgot to write a test for it?

alexmamonchik: I wanted to add it in another commit because that isn't refactoring, that's new fix/feature.

PR #237, opened 2015-06-06
3.x example, live tests uses rules, not keys

kathyonu:
Added the 3.x live tests example code.


mindeavor:
Thank you! Would you mind squashing these into a single commit?

PR #245, opened 2015-07-08
rename card to source in README.md

tzzzoz:
I was new to stripe-ruby-mock.
When I tried to follow this README, I found this problem.
In the beginning, I found that I couldn't retrieve the card which I created for a new customer. I was confused with this, and finally I found this come from README.
I hope it could be helpful for some new guys.

Cheers.


bhaity:
+1 for this

alexmamonchik:
@tzzzoz test has been failed. Please, fix the error and update PR. @bhaity you can help with that

PR #252, opened 2015-08-13
Allow accounts to be created and then fetched by id

johanoskarsson:


johanoskarsson:
Any thoughts on this one?

alexmamonchik:
@johanoskarsson please, add `live: true` to your test. I tried and test was failed.

johanoskarsson:
I attempted to make the test live compatible but unfortunately I'm not sure it's possible right now.
Since stripe only allows creation of one account per email address (and per api key?) I need to be able to delete the previous one to create a new one. However stripe account delete is not in the stripe gem used by stripe-ruby-mock. They only added that in July it seems.

Alternatively perhaps I could try to generate some sort of unique email address on each test run, but that feels janky and will probably cause flaky tests.

Any ideas?

alexmamonchik:
Yes, I tried create tests for accounts too, but this is not so easy. Also, It's a part of Stripe Connect - not a basic Stripe functionality. Maybe we don't need to include it here. You can use `managed` accounts - it doesn't require unique email.

alexmamonchik:
also look at #218

johanoskarsson:
I'm a bit confused by the merged commit msg above. I don't see cc26fe2 reflected in master. Did this get merged somewhere else?

PR #257, opened 2015-09-12
Cardfix

kathyonu:
Fixed incorrect card code, updates hash format, formated with rubocop specs.


louim:
I think it would be better to separate style fixes from bug fixes.

Maybe create another pull request with the styles fixes, and include the rubocop file so people can run it themselves? It could also be setup to run on the CI (via [HoundCI](https://houndci.com/)).

You should squash the commits.

kathyonu:
@louim Thank You .. I am currently studying how to best revert / fix that failing commit .. I have a question in to @alexmamonchik on this problem, and what I think is a best solution.  I share it here:
```ruby
git stash # saves my current work in my stripe-1.24.0 branch
git checkout cc0f8cf # checks out the breaking commit
git checkout -b fixcc0f8cf # create a new branch from it, make changes in new branch
git commit -m 'fixes ccof8cf' # commits changes in the new branch
git push origin fixcc0f8cf # Issue a PR on this branch, this should fix that broken commit 
git checkout stripe-1.24.0
git stash apply
```
I am still studying this, however ..
Any thoughts you have on how to best fix that failing commit, I am all ears.  
I like your idea about a separate PR for rubocop changes, and including it so folks can run it. ( it is entirely overwhelming the first time you run it on the entire gem. you have to run it file by file ) I have done that in my fork's stripe-1.24.0 branch, which is where I am working to bring our stripe-ruby-mock gem up to stripe-1.25.0.  When I have the 1.24.0 branch ready to be called stripe-1.25.0, i will do a squash commit into a new branch named stripe-1.25.0 .  It will have any and all files I changed in any way structurally/code wise, to mirror stripe itself, also changed to the new hash form, and changed to conform as close as possible to rubocop.  I am ending up with much cleaner, clearer and more readable code from these changes.  From this new master, I don't know what will be the best way to bring all those changes into the gem.  Right now, it affects a lot of files, but nowhere near all.

I also just included the simplecov gem, which gives a great picture of the code test coverage.

PR #268, opened 2015-10-12
Add a method for deauthorizing accounts

gregawoods:
This request adds a handler for the `Account#deauthorize` endpoint. I needed to test this endpoint in my app, which is using Stripe Connect.

Note: Technically the deauthorize method hits `/oauth/` instead of `/v1/accounts`, so I could see an argument for moving this code into another module. On the other hand, the official Stripe gem actually lumps this functionality into the `Account` class with the other account related api calls. So I chose to follow Stripe's example and do the same.

### DIFF
--- a/lib/stripe_mock/request_handlers/accounts.rb
+++ b/lib/stripe_mock/request_handlers/accounts.rb
@@ -8,6 +8,7 @@ module StripeMock
         klass.add_handler 'get /v1/accounts/(.*)',  :get_account
         klass.add_handler 'post /v1/accounts/(.*)', :update_account
         klass.add_handler 'get /v1/accounts',       :list_accounts
+        klass.add_handler 'post /oauth/deauthorize',:deauthorize
       end
 
       def new_account(route, method_url, params, headers)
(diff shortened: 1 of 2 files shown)


alexmamonchik:
I have small updates in master branch. Please, update your PR.

gregawoods:
Ah, I see where this is failing now. I didn't realize that the tests also ran against the live stripe servers. My test code is using a dummy client_id value, so that is going to fail.

I suspect that this will be pretty difficult to run live, considering it requires not only a real client_id but also a connected account in order to achieve the desired response. I'm open to suggestions though.

alexmamonchik:
I can't find endpoint in Stripe doc for that. Can you explane API version and case how do you test it

gregawoods:
Hi Alex. This is part of the Stripe Connect API and can be found here: https://stripe.com/docs/connect/reference#post-deauthorize

You would use this in an application that uses Stripe Connect to let users authenticate their Stripe accounts. If the user decides they no longer want to use Stripe with your app, you call the deauthorize API to revoke the access token. You can think of this as "logging out" the users Stripe account from your app.

PR #280, opened 2015-11-29
Prevent metadata being lost on charge update

ball-hayden:
Currently, updating metadata on a charge causes existing metadata (and some other stored data, such as refunds) to be lost.

This appears to be because we call `merge`, while really we should be recursively merging the existing charge hash with the new parameters.

This PR updates the `update_charge` method to use `rmerge`, and adds a simple test case.


kathyonu:
Please change this: it "does not loose data when updating a charge" do
To this: it "does not lose data when updating a charge" do
loose to lose.
and :+1:

ball-hayden:
Ooops. Thanks :flushed:

PR #282, opened 2015-12-01
Charge requires either source or customer to be set

jclay:
As per the Stripe docs, [creating a new charge](https://stripe.com/docs/api/ruby#create_charge) requires either a customer token or a source to be set:
screen shot 2015-12-01 at 5 43 40 pm


alexmamonchik:
Hi, Joel! Thanks for contributing! Please, fix your code and update your branch from master

PR #283, opened 2015-12-03
Fix for saving a customer invalidating sources

benjie:
If you have a customer who has a source on the server, you pull it down, and then you call save, the Stripe Ruby gem sends up `sources: [{}]` where I'm guessing the `{}` is a placeholder to say "don't change this source" (seems a bit weird to me - surely at least including an id would be a good idea?). I think it has something to do with [these lines from the stripe ruby gem](https://github.com/stripe/stripe-ruby/blob/4e01b13efcc6557ca0f78d20e93f4b7c617d94a3/lib/stripe/stripe_object.rb#L146-L148). Or maybe something else is going on... I don't know.

Anyway, when we wanted to just update a `metadata` key on the customer, this was replacing the cards with empty hashes and that meant even doing `card.id` started throwing errors. This patch fixes it by just copying the old data with the relevant index from the previous data in the case of empty hashes.

Let me know if you'd like me to make any tweaks, this is my first Ruby pull request so very happy to receive guidance!

### DIFF
--- a/lib/stripe_mock/request_handlers/customers.rb
+++ b/lib/stripe_mock/request_handlers/customers.rb
@@ -62,6 +62,10 @@ module StripeMock
         # Delete those params if their value is nil. Workaround of the problematic way Stripe serialize objects
         params.delete(:sources) if params[:sources] && params[:sources][:data].nil?
         params.delete(:subscriptions) if params[:subscriptions] && params[:subscriptions][:data].nil?
+        if params[:sources] && !params[:sources][:data].nil?
+          # Copy old sources over empty objects
+          params[:sources][:data].collect!.with_index {|v, i| v.keys.length == 0 ? cus[:sources][:data][i].clone : v }
+        end
         cus.merge!(params)
 
         if params[:source] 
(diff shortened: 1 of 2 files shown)


benjie:
Turns out the same issue occurs for subscriptions too, only in the case of subscriptions the objects sent are: `{:sources=>{:data=>[{}]}, :subscriptions=>{:data=>[{:plan=>{}, :metadata=>{}}]}}` so I've switched to assuming that they must have a `type` to be valid.

benjie:
... And of course this breaks `total_count`; so I've switched to just deleting the entire params unless it contains something that looks valid.

(Otherwise if you try and cancel the subscription it throws:

```
#<NoMethodError: undefined method `-' for nil:NilClass>
.../bundler/gems/stripe-ruby-mock-454a71c71cc6/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb:37:in `delete_subscription_from_customer'
```
)

mattshaffer11:
Would love to get this merged :+1:

alexmamonchik:
great job! Thanks

PR #292, opened 2015-12-27
Added missing fields to mocked invoices

TrevorHinesley:


TrevorHinesley:
Added some missing fields to mocked charges as well.

Issue #219, opened 2015-04-25
StripeMock.spawn_server throws an error when passing "-e" option to RSpec.


georgemillo:
RSpec has a command line option "-e" or "--example" that lets you filter tests by their description.

    $ rspec -r "creates a new user"
    # => This will run the test defined by 'it "creates a new user"'.

But with stripe-ruby-mock installed, trying to run rspec with the `-e` flag throws an error:

    $ rspec --example "creates a new user" spec/features/users/sign_up_spec.rb
    /Users/george/.rvm/gems/ruby-2.2.1/gems/dante-0.2.0/lib/dante/runner.rb:212:in `parse_options': invalid option: --example (OptionParser::InvalidOption)
	from /Users/george/.rvm/gems/ruby-2.2.1/gems/dante-0.2.0/lib/dante/runner.rb:50:in `execute'
	from /Users/george/.rvm/gems/ruby-2.2.1/gems/stripe-ruby-mock-2.1.1/lib/stripe_mock/api/server.rb:19:in `spawn_server'
	from /Users/george/path/to/my/app/spec/spec_helper.rb:11:in `<top (required)>'

The offending line, spec_helper:11, is "StripeMock.spawn_server". Commenting that line out makes the `-e` flag work as intended, but then of course my tests don't work because the StripeMock server isn't running.

Looks like the problem is being caused by the Dante gem but I've never used that and don't know what it does.

kathyonu:
May I recommend you reinstall Dante .. I had a different problem, and doing this fixed it .. 
I trashed my Gemfile.lock, simply move it to the trash.  Doing so gives your bundle a change to cleanly reset all the dependencies. Then ..

$ gem uninstall dante
$ bundle
$ gem install dante
$ bundle update

It installed the exact same version, and yet doing this worked.
I can't explain it, I just know it worked for me.
Hope it works for you.

georgemillo:
Tried your exact steps but it didn't fix the problem.

For the record, here's my `Gemfile.lock`:

```
GEM
  remote: https://rubygems.org/
  specs:
    actionmailer (4.2.1)
      actionpack (= 4.2.1)
      actionview (= 4.2.1)
      activejob (= 4.2.1)
      mail (~> 2.5, >= 2.5.4)
      rails-dom-testing (~> 1.0, >= 1.0.5)
    actionpack (4.2.1)
      actionview (= 4.2.1)
      activesupport (= 4.2.1)
      rack (~> 1.6)
[...]

PR #295, opened 2016-01-14
Allow arguments when spawning a server

fleaplus:
## What does this pull request do?

Allows arguments to be used and be preserved when spawning a StripeMock server.
## Can you provide any background information about this?

I ran into this issue when attempting to override the output format in rspec. The spawned server did not understand the `--format documentation` argument and threw an exception.

Provided this fix looks ok, its a potential candidate for an upstream PR to address #219 
## Does this change specified behavior (did you have to change tests), and why?

I added an extra context into the `spec/server_spec.rb` file so I could modify ARGV before spawning the server in the test.
## Definition of Done:
- [x] Code Tested?
- [ ] Migrations?
- [ ] Readme Updated?
- [ ] Additional Dependencies?


alexmamonchik:
@fleaplus really thanks for reviewing this issue. But looks like your PR has a lot of problems, I can't merge it. But if you fix it I'll merge it.

PR #307, opened 2016-02-11
Added `currency` attribute to customer

slash7:
Adds `currency` attribute to customer.
https://stripe.com/docs/api?lang=ruby#customers


alexmamonchik:
@slash7 please, add test for your changes https://github.com/rebelidealist/stripe-ruby-mock/wiki/Implementing-a-New-Behavior . And thanks for your help.

slash7:
@alexmamonchik PTAL

PR #315, opened 2016-03-10
fix stripe

OlenaOhranchuk:


alexmamonchik:
because https://github.com/rebelidealist/stripe-ruby-mock/commit/64b3e94c3b3d826859578aea289bd726aa6c8db0 . Not all users want to use last stripe and not all can do that. Now If you want 1.36. or more - define it in your Gemfile

alexmamonchik:
but if you can help with some other issues - welcome! thanks

Issue #209, opened 2015-03-30
Nil api key causing exception with Stripe 1.20.4


crberube:
Wanted to point out this issue that I noticed even though stripe-ruby-mock is only supporting stripe 1.20.1 as of now.

As of the following PR https://github.com/stripe/stripe-ruby/pull/221 stripe raises a TypeError when it receives an api_key in the options that is not a string. In the mock_request method of StripeMock::Instance, api_key is a required parameter that is nil by default, thus raising an exception unless an api key is explicitly passed.

As an example, the following code raises a TypeError:
```ruby
stripe_customer.sources.create(card: stripe_token)
```

while this code works as expected:
```ruby
stripe_customer.sources.create({card: stripe_token}, {api_key: Stripe.api_key})
```

DanielePalombo:
same problem

aliou:
Having the same problem.

jasonrhalpern:
I think I have a somewhat similar problem. In various methods in my code I use things like: 

customer.subscriptions.retrieve({SUB_ID})
customer.sources.retrieve({CARD_ID})

I am using this gem to test the methods that then make these API calls to Stripe. The ones for create customer/subscription/plan are working fine without the api_key, but all of the calls to the retrieve methods (listed above) are failing with the following error:

TypeError: api_key must be a string.

micahwedemeyer:
+1 same problem

jb41:
Anyone found solution to that?

jasonrhalpern:
I still haven't figured out a solution to that issue.

samkpower:
+1

psiedlak:
+1

moneytree-doug:
:+1:

justinbburris:
@jasonrhalpern

Similar to what @crberube mentioned, you can (for now) make your call with the api_key

```ruby
stripe_customer.sources.retrieve(stripe_card_id, {api_key: Stripe.api_key})
```

alexmamonchik:
you can use branch https://github.com/rebelidealist/stripe-ruby-mock/tree/stripe-1.21.0

jasonrhalpern:
Thanks for the help on this guys, I'll probably use the older branch for the time being. I don't really want to have to modify the stripe code that I am testing by adding the api_key since doing so would only be necessary for the tests to pass and not for the actual implementation.

moneytree-doug:
[...]

PR #322, opened 2016-04-14
Avoid to require an api key to get the instance working

lucas-aragno:
related to this issue #209 

It seems pointless to me to need to add a api_key that it's not being used for almost all ( at least the functions that i need) the functions from the stripe instance so this will be a patch for this, you can still set up the api key if you want but if you don't you are good to go w/ the dummy


chibicode:
@alexmamonchik could you bump the version?

alexmamonchik:
@chibicode ask @mindeavor

maccman:
Please release this

mindeavor:
Version was bumped in this commit: a37f724e5c8a72482b7294bf1a6b0aa6a1eed50a

PR #325, opened 2016-05-03
Implement basic support for bank account sources

lfittl:
This allows creating a customer using bank account tokens, defaulting to "new" state, and then always marking as verified when verify endpoint gets called.

---

Note that this was the minimum change necessary in order to make our feature specs work - I'm sure one could cover the API more extensively, but this is the minimum so we can use this library for mocking.


lfittl:
@alexmamonchik Yay, thanks for the fast merge - that makes me eager to contribute more in the future 🎉

alexmamonchik:
@lfittl You're welcome! I started as you. Just sent PR with changes what I wanted. This is really nice library, let's make it more powerful!

PR #327, opened 2016-05-04
Adds remote error queue and replaces Jimson with Ruby's DRb

Willianvdv:
This adds error sharing between clients. This is useful when you want to prepare an error on client A that lets a request on client B fail. In our case we have an "external" service that's using stripe ruby mock and an another service that runs the specs and prepares and Stripe::Card error.

I've replaced Jimson with Ruby's [DRb](http://ruby-doc.org/stdlib-1.9.3/libdoc/drb/rdoc/DRb.html) so actual Ruby objects can be shared, in this case the prepared error type. Before switching to DRb I tried to implement this in Jimson but I didn't like the serialisation/marshalling of error types too much. In this application, where client and server share the same code, DRb looked like an ideal candidate to share client / server data.

This pr still needs some TLC.

### DIFF
--- a/lib/stripe_mock/api/client.rb
+++ b/lib/stripe_mock/api/client.rb
@@ -30,8 +30,8 @@ module StripeMock
   def self.redirect_to_mock_server(method, url, api_key, params={}, headers={}, api_base_url=nil)
     handler = Instance.handler_for_method_url("#{method} #{url}")
 
-    if mock_error = client.error_for_handler_name(handler[:name])
-      client.dequeue_error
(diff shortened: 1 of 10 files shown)


pwim:
Being able to drop Jimson would be awesome, as it looks to be no longer maintained, and is causing this gem to not work under Rails 5 (see #341). You mention this PR needs some TLC. Anything specific?

Willianvdv:
Remove the leftover Jimson stuff and remove it as dependency. I'll create a cleanup pr!

pwim:
@Willianvdv Good to know! I've already created a clean up PR in #342.

PR #329, opened 2016-05-05
Master Blaster: Fix commits lost by force push, lock Stripe to < 1.42

typeoneerror:
* Reintroduces master commits deleted by force push.
* Locks Stripe to < 1.42 to avoid failing tests with new subscriptions API introduced in 1.42 (https://github.com/stripe/stripe-ruby/blob/master/History.txt).

See https://github.com/rebelidealist/stripe-ruby-mock/pull/326 for the rabbit hole I went down with this one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This gem has unexpectedly grown in popularity and I've gotten pretty busy, so I'
 
 In your gemfile:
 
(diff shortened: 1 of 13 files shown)


typeoneerror:
cc @mindeavor @alexmamonchik

alexmamonchik:
:clap: thanks a lot @typeoneerror

PR #338, opened 2016-06-01
Support generate_bank_token in stripe_helper

Empact:
This mostly adapts existing card flows to more general sources flows, and
adds bank account support along the way.

This is not a complete solution, but it improves support significantly.


Empact:
Test failures look to be unrelated.

PR #343, opened 2016-06-14
Update to 1.43

adbjesus:
Deals with issue #335 

I guess we should bump the version. But i am not sure if it should be to 2.2.5 or 2.3.0. So I will wait for feedback before doing that.

PR #344, opened 2016-06-22
Support old-style URLs

pedantic-git:
Stripe has changed its URLs for subscriptions.

The Ruby library still uses old-style Stripe URLs in some situations at least (possibly all situations?).

This changes the regexps so that the mock server responds whichever style of URL is used. I found I needed this change to get subscriptions working.

Additionally, the customer regexps are tightened so they're not accidentally mistaken for subscription regexps.

### DIFF
--- a/lib/stripe_mock/request_handlers/customers.rb
+++ b/lib/stripe_mock/request_handlers/customers.rb
@@ -4,9 +4,9 @@ module StripeMock
 
       def Customers.included(klass)
         klass.add_handler 'post /v1/customers',                     :new_customer
-        klass.add_handler 'post /v1/customers/(.*)',                :update_customer
-        klass.add_handler 'get /v1/customers/(.*)',                 :get_customer
-        klass.add_handler 'delete /v1/customers/(.*)',              :delete_customer
+        klass.add_handler 'post /v1/customers/([^/]*)',             :update_customer
+        klass.add_handler 'get /v1/customers/([^/]*)',              :get_customer
+        klass.add_handler 'delete /v1/customers/([^/]*)',           :delete_customer
         klass.add_handler 'get /v1/customers',                      :list_customers
       end
 
--- a/lib/stripe_mock/request_handlers/subscriptions.rb
+++ b/lib/stripe_mock/request_handlers/subscriptions.rb
@@ -3,11 +3,11 @@ module StripeMock
     module Subscriptions
 
       def Subscriptions.included(klass)
-        klass.add_handler 'get /v1/customers/(.*)/subscriptions', :retrieve_subscriptions
-        klass.add_handler 'post /v1/customers/(.*)/subscriptions', :create_subscription
-        klass.add_handler 'get /v1/customers/(.*)/subscriptions/(.*)', :retrieve_subscription
-        klass.add_handler 'post /v1/customers/(.*)/subscriptions/(.*)', :update_subscription
-        klass.add_handler 'delete /v1/customers/(.*)/subscriptions/(.*)', :cancel_subscription
+        klass.add_handler 'get /v1/customers/(.*)/subscription(?:s)?', :retrieve_subscriptions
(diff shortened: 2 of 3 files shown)


alexmamonchik:
Already implemented in #343

pedantic-git:
Are you sure this is implemented in #343? This re-adds support for the `/subscription` URLs (rather than `/subscriptions`). At the time of this PR, the `stripe` gem was still trying to call those URLs, and the upstream Stripe server was still responding to them.

alexmamonchik:
Maybe wrong PR but tested URL like 'get /v1/customers/(.*)/subscription/(.*)' and got
```bash
{
  "error": {
    "type": "invalid_request_error",
    "message": "Unrecognized request URL (GET: /v1/customers/cus_8Sm78WNNk0zk6z/subscription/sub_8Sm7pLjYrLIlhZ). Please see https://stripe.com/docs or we can help at https://support.stripe.com/."
  }
}
```
right URL is 'get /v1/customers/(.*)/subscriptionS/(.*)' but check https://github.com/rebelidealist/stripe-ruby-mock/blob/master/lib/stripe_mock/request_handlers/subscriptions.rb#L13 If you think we still need some routes - just send one more PR I'll accept it

pedantic-git:
Here's a simple spec that illustrates the issue (this is using `stripe` 1.50.0 and `stripe-ruby-mock` 2.3.1):

```ruby
  describe 'stripe test' do
    it 'works with the latest stripe-ruby-mock gem' do
      token = StripeMock.create_test_helper.generate_card_token
      cus   = Stripe::Customer.create(source: token)
      expect {
        cus.update_subscription(plan: 'default')
      }.not_to raise_error
    end
  end
```

```
Failures:

  1) Subscription stripe test works with the latest stripe-ruby-mock gem
     Failure/Error:
       expect {
         cus.update_subscription(plan: 'default')
       }.not_to raise_error
     
       expected no Exception, got #<Stripe::InvalidRequestError: (Status 404) No such customer: test_cus_3/subscription> with backtrace:
         # /bundle/gems/stripe-ruby-mock-2.3.1/lib/stripe_mock/instance.rb:117:in `assert_existence'
         # /bundle/gems/stripe-ruby-mock-2.3.1/lib/stripe_mock/request_handlers/customers.rb:61:in `update_customer'
         # /bundle/gems/stripe-ruby-mock-2.3.1/lib/stripe_mock/instance.rb:96:in `mock_request'
         # /bundle/gems/stripe-ruby-mock-2.3.1/lib/stripe_mock/api/instance.rb:22:in `call'
         # /bundle/gems/stripe-ruby-mock-2.3.1/lib/stripe_mock/api/instance.rb:22:in `block in alias_stripe_method'
         # /bundle/gems/stripe-1.50.0/lib/stripe/api_operations/request.rb:15:in `request'
         # /bundle/gems/stripe-1.50.0/lib/stripe/api_operations/request.rb:37:in `request'
         # /bundle/gems/stripe-1.50.0/lib/stripe/customer.rb:62:in `update_subscription'
[snip]
```

With my changes, this spec doesn't fail. Does it work for you? I did have `Stripe.api_version = '2015-02-16'` but even with this commented out I get the same failure.

Thanks for your help!

pedantic-git:
Hi @alexmamonchik - can I please check in about this issue? I know the PR is still closed but hopefully my spec above illustrates it's still an issue.

alexmamonchik:
Hello. Let's review it again. Did you add the example test to PR?

pedantic-git:
@alexmamonchik - I didn't, but I will do later this week. Thanks for looking at it again.

pedantic-git:
@alexmamonchik I've added a new spec that tests the update_subscription method of customers. It fails on the current master and passes with this PR.

There's another spec failure above but it looks like it's because one of Stripe's exception messages has changed and isn't related to this PR.

PR #349, opened 2016-06-29
simple support to stripe 1.43.1

caimano:


brianlittmann:
`stripe-ruby` has already been bumped to 1.44.0. I still suggest we drop the max version cap and just leave it as `gem.add_dependency 'stripe', ['>= 1.31.0']` unless that would open a can of worms I'm unaware of.

Rabsztok:
I agree, version 1.43 of `ruby-stripe` is unusable in my application. I get tons of `no implicit conversion of Symbol into Integer` errors. Using more recent versions of `ruby-stripe` helped.

alexmamonchik:
unlocked

PR #350, opened 2016-07-01
Check for `trial_end` on a subscription before raising error that card is required

cannikin:
The current check as to whether adding/modifying a subscription requires a card be present only checks the `plan` and `params` for a `trial_end` date, but not the subscription object itself. The [Stripe subscription docs](https://stripe.com/docs/api#create_subscription-trial_end) state you can set the `trial_end` date on the subscription object itself which will override any `trial_end` set on a plan. I'm using this functionality in my app and updating to version 2.3.0 of stripe-ruby-mock broke one of my tests which updates an existing subscription to a new one, both with a 3-day trial. Stripe itself allows this to happen without a card being present.

This PR updates the `verify_card_present` method to require the subscription object as a parameter and checks its `:trial_end` key before raising that "card missing" error.

The specs pass as-is so apparently this check isn't currently being tested. I wanted to open this PR first to be sure this is a change that makes sense and would be accepted. If so I'll go in and add those missing specs.


vadzimpiatkevich:
A build fails due to an externality. The source is actually correct.
Build was fixed in #352 
So, you could close the PR then open it again. This will trigger a new build. Or @alexmamonchik could "Restart Build" on the build's detail screen.

alexmamonchik:
@cannikin thanks for improving. I'm going to merge this PR. Can you add tests related to your changes?

cannikin:
@alexmamonchik Hopefully I'll have time this week to add some tests around this change, thanks for taking a look!

alexmamonchik:
@cannikin Let's cofirm. 
```ruby
customer = Stripe::Customer.create(id: 'blabla')
plan = stripe_helper.create_plan(id: 'free', amount: 100, trial_period_days: 14)
sub = customer.subscriptions.create(plan: plan.id)
# sub.trial_end should be today + 14 days

sub.trial_end = Time.now.to_i + 36000
sub.save
# now sub.trial_end should be now + 10 hours
```

PR #355, opened 2016-07-11
Add support for delete_discount, along with discount#start and discount#end

andyzei:
Add support for `subscription#delete_discount`, `discount#start`, `discount#end`


andyzei:
need to fix tests

Issue #364, opened 2016-08-15
Issue when deleting coupon from subscriptions


rajahafify:
So I have a subscription with discount

```
#previously
stripe_subscription.coupon = subscription.coupon.code
```

So when I try to remove the discount by attaching an empty string 
`stripe_subscription.coupon = ""`
stripe respond with this error
`You cannot set coupon to an empty string. We interpret empty strings as nil in requests. You may set (object).coupon = nil to delete the property.`

```
Failure/Error: stripe_subscription.coupon = ""

     ArgumentError:
       You cannot set coupon to an empty string. We interpret empty strings as nil in requests. You may set (object).coupon = nil to delete the property.
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-1.42.0/lib/stripe/stripe_object.rb:233:in `block (3 levels) in add_accessors'
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-1.42.0/lib/stripe/stripe_object.rb:264:in `call'
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-1.42.0/lib/stripe/stripe_object.rb:264:in `method_missing'
```

But when i use nil instead `stripe_subscription.coupon = nil`, stripe mock complaint that the coupon doesn't exist
```
Failure/Error: stripe_subscription.save

     Stripe::InvalidRequestError:
       No such coupon:
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-ruby-mock-2.3.0/lib/stripe_mock/request_handlers/subscriptions.rb:146:in `update_subscription'
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-ruby-mock-2.3.0/lib/stripe_mock/instance.rb:96:in `mock_request'
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-ruby-mock-2.3.0/lib/stripe_mock/api/instance.rb:22:in `call'
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-ruby-mock-2.3.0/lib/stripe_mock/api/instance.rb:22:in `block in alias_stripe_method'
     # /Users/rajahafify/.gem/ruby/2.1.4/gems/stripe-1.42.0/lib/stripe/api_operations/request.rb:15:in `request'
[...]

PR #365, opened 2016-08-15
Allow nil to be passed to StripeSubscription object as per the API during update_subscription

rajahafify:
Possible fix for #364


rajahafify:
I've added a test suite to make sure im not breaking anything

PR #368, opened 2016-08-28
Add Country Spec Endpoint

luaybs:
This PR adds the following endpoint from the Stripe API:

https://stripe.com/docs/api#retrieve_country_spec

The US country spec is returned as mock data.

Let me know what you think. :)


stripe-ruby-mock.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.require_paths = ['lib']
 
-  gem.add_dependency 'stripe', '>=1.31.0'
+  gem.add_dependency 'stripe', '>=1.40.0'
```
alexmamonchik: I don't think that we can do that now. A lot of users still use old version of stripe. We can update it later

alexmamonchik:
Hi. Thanks for help. I like it except stripe >= 1.40. I'll merge PR without commit  	8d528af

luaybs:
Thanks I think I replaced that line with what's on master when I merged it in here:
https://github.com/rebelidealist/stripe-ruby-mock/pull/368/commits/0104becefb5fd0eca00ef2f07e4e992ea5e59578#diff-ebf1b65abf5e55397a399fae9331f0e8R20

PR #380, opened 2016-10-09
Support for application fees and generation of invoices for subscription renewals

hoyaemt:
We use Stripe extensively with Connected (ie, managed) accounts, monthly subscriptions, and application fees and need some additional functionality added to this excellent mocking gem to fully support our testing.  This pull request includes the following:  
- Add support for Stripe application fees - which are only applicable to charges made in managed accounts.  Application fees are now created for these charges.  A balance transaction is also created for the application fee.
- Add test helper method (generate_subscription_renewal_invoice) which accepts a Stripe subscription ID and generates the appropriate invoice which would be created when the subscription is renewed at the beginning of a period.
- Update existing code that handles balance transactions to calculate the net amount programatically based on the supplied amount and stripe_fee_amount.  If the charge includes an application fee, the net amount reflects the subtraction of the application fee.
- A balance transaction is not created when a charge is created with capture=false.
- When stripe_account is specified in the API header (which indicates activity for an item in a connect account) for a new charge or customer, this account number is persisted.
- Update the pay_invoice method of Stripe invoices to create the related charge, balance transaction, and application fee (if necessary) instead of setting the invoice's charge to a non-existent Charge ID.
- Add support for creating credit card and bank account tokens in managed accounts.

### DIFF
--- a/lib/stripe_mock.rb
+++ b/lib/stripe_mock.rb
@@ -44,6 +44,7 @@ require 'stripe_mock/request_handlers/helpers/token_helpers.rb'
 require 'stripe_mock/request_handlers/validators/param_validators.rb'
 
 require 'stripe_mock/request_handlers/accounts.rb'
+require 'stripe_mock/request_handlers/application_fees.rb'
 require 'stripe_mock/request_handlers/balance_transactions.rb'
 require 'stripe_mock/request_handlers/charges.rb'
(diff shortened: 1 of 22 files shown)


alexmamonchik:
Hi. Thanks for your help. I can't accept this PR. Please, split it to few PRs and merge commits to 1-2

Issue #310, opened 2016-03-02
NoMethodError: undefined method `external_accounts' for #<Stripe::Account:0x007fc131029168>


ryanfelton:
For Stripe Managed Accounts, they now have the associated object "external accounts" that contain bank routing and bank account numbers.

https://stripe.com/docs/api#account_object-external_accounts

I would love to see external accounts supported by stripe-ruby-mock!

Aeradriel:
It seems not to be only external accounts but all managed accounts. There is no `verification` field either.

abmxer:
Any thoughts on how to handle this?

Aeradriel:
If you are testing web hooks, you can create a json file containing a sample response with the name of your web hook (i.e. account.external_account.created) in spec/fixtures/stripe_web_hook. It will be automatically loaded by the gem.

I you want to test basics calls (such as `Stripe::Customer.create`), I do not think it is implemented yet. I will maybe contribute to the gem to fix it this week end if I can find some time. :smile:

abmxer:
Actually, I'm trying to create an bank account. 

```
  1) StripeService Customer should update a managed account
     Failure/Error: account.external_accounts.create(:external_account => options)
     
     NoMethodError:
       undefined method `external_accounts' for #<Stripe::Account:0x0055ed8c2a17d0>
     # /usr/local/bundle/gems/stripe-1.50.0/lib/stripe/stripe_object.rb:271:in `method_missing'

```

I tried to redefine the ```external_accounts``` inside my test, but I was not able to. 

```
      module StripeMock
        def external_accounts=
        end

        def external_accounts
        end
      end
      Stripe::Account.extend StripeMock
```

Any thoughts on how to monkey patch this until there be an official support? 

Thanks!

Aeradriel:
Well, I made a PR that was just merged to add external_accounts to account's hash and external_accounts webhooks.
Adding the whole external accounts logic will take a little more time considering the fact that I need to dive deeper into the code to fully understand the behaviour of the gem. I will start working on it tomorrow and I will notify you here when I'll be done with it. :)

PR #383, opened 2016-10-20
Add External Accounts

davidstump:
This commit adds the ability to test external accounts for Managed
Accounts. Managed Accounts can have external accounts (very similar to
sources for Customers) that can be either a bank account or a debit
card. These changes include all of the mocks, specs, helpers and logic
needed to start testing these external accounts.


davidstump:
There had been some previous work done on External Accounts (see: #310), however, there were no request handers and other logic in place to more fully support testing Managed Accounts with external bank accounts.

davidstump:
Hi Alex. Just saw this comment from Jan. Can I confirm that the test account belongs to @mindeavor that needs the adjustment? I re-ran the specs locally and merged with upstream just to make sure nothing seemed off with the initial PR. Thanks :)

mindeavor:
I don't have my account activated, but if someone wants to PM me some test keys on gitter, I can update the test setup to use them.

arthurnn:
so, what is holding this off?

Dashue:
I need this naaaooooo, thanks :)

davidstump:
I believe the hold-up at the moment is the test account that was being used needs to be verified in order to be able to deal with external accounts.

Dashue:
Cheers! Will be looking to utilise this as soon as it's in.

davidstump:
@alexmamonchik Are there any other stripe test accounts that you have used that could be activated?

Dashue:
Words cannot express how much I need this. Maybe reach out to stripe to get sponsored with a dedicated account for this project?

alexmamonchik:
the problem is api version + stripe test keys. So, stripe-ruby-mock follows 2015-09-08 api version. We can't register platform on existing keys because because we don't have access to those accounts. We can change keys to new, but I can create accounts with 2017-02-14 api version - it breaks Recipient feature. I suggest to merge this PR to new branch and new minor version 2.5.0. Also, include #410 to this new 2.5.0 version. What do you think?

Dashue:
I wouldn't be able to comment in regards to #410, sounds like it would be good though?
If there's anything I can chip in with, please let me know.

Dashue:
To be honest, what ever simplest thing that get it's working the fastest. This has been open since oct, and a lot of people would find this functionality useful. I personally don't care if some automated tests aren't working due to problems with api keys. If this won't be sorted I'll probably just pick the changes from this changeset myself to get it working.

Dashue:
As an update to this, I pulled in this PR locally and it worked fine for my tests. We're fine using our own version of stripe-ruby-mock until this gets fixed.

borenho:
@Dashue , @davidstump @alexmamonchik ,  what is the way forward for this feature? I bet many people need this working

PR #389, opened 2016-11-06
Add validations to Charge API

knjko:
```ruby
describe '#charge' do
  subject { Stripe::Charge.create(params) }
  let(:params) { { amount: 10, currency: 'usd' } }

  before { StripeMock.start }
  after { StripeMock.stop }

  it { expect { subject }.to raise_error(Stripe::InvalidRequestError) }
end

# => expected Stripe::InvalidRequestError but nothing was raised 
```

```ruby
Stripe::Charge.create({amount: 10, currency: 'usd'})
# => Stripe::InvalidRequestError: Must provide source or customer.
```

- [Either `source` or `customer` is required](https://stripe.com/docs/api?lang=ruby#create_charge-customer) in Charge API, so I added validations.
- There are specs passes `card` param to Charge API. But the API is [not allowed](https://stripe.com/docs/api?lang=ruby#create_charge) `card`, so I move from `card` to `source` param in almost specs.


knjko:
👍

PR #394, opened 2016-11-27
Allow for charge creation with bank token

golmansax:
Not sure when this change was made live in the API, but it's backwards compatible.  Let me know what you think!


golmansax:
Thanks!

PR #405, opened 2017-01-11
Expand balance transaction when retrieving an existing charge

vkinelev:
- `Stripe::Charge.retrieve` includes balance transaction object if `expand: ['balance_transaction']` has been passed into params. 
- Added spec

### DIFF
--- a/lib/stripe_mock/request_handlers/charges.rb
+++ b/lib/stripe_mock/request_handlers/charges.rb
@@ -49,12 +49,13 @@ module StripeMock
             params.merge :id => id,
             :balance_transaction => balance_transaction_id)
 
+        charge = charges[id].clone
         if params[:expand] == ['balance_transaction']
-          charges[id][:balance_transaction] =
+          charge[:balance_transaction] =
             balance_transactions[balance_transaction_id]
         end
 
-        charges[id]
+        charge
       end
 
       def update_charge(route, method_url, params, headers)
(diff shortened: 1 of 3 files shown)


vkinelev:
I'm trying to understand why Travis build fails. I haven't touched that code:

```
StripeMock::Instance behaves like Coupon API retrieve coupon cannot retrieve a stripe coupon that doesn't exist
     Failure/Error: expect { Stripe::Coupon.retrieve('nope') }.to raise_error {|e|
       
       expected: "id"
            got: "coupon"
       
       (compared using ==)
     Shared Example Group: "Coupon API" called from ./spec/support/stripe_examples.rb:15
     # ./spec/shared_stripe_examples/coupon_examples.rb:57:in `block (3 levels) in <top (required)>'
```

A spec is:

```ruby
    it "cannot retrieve a stripe coupon that doesn't exist" do
      expect { Stripe::Coupon.retrieve('nope') }.to raise_error {|e|
        expect(e).to be_a Stripe::InvalidRequestError
        expect(e.param).to eq('id')
        expect(e.http_status).to eq(404)
      }
    end
```

Live Stripe returns "coupon" type in InvalidRequestError, but mock server returns "id":
```ruby
      def get_coupon(route, method_url, params, headers)
        route =~ method_url
        assert_existence :id, $1, coupons[$1]
      end
```

I probably have to setup live testing locally :)

vkinelev:
I got the same error when I run tests locally. It happens both in my branch and commit tagged with "v2.4.0".

A hint would be so much appreciated! :)

psusmars:
I also received a similar response from Travis when submitting a pull request that had nothing to do with the coupon code.

alexmamonchik:

vkinelev:
@alexmamonchik done!

Issue #406, opened 2017-01-17
Pagination with "starting_after" in an all call does not act as it does on Stripe


psusmars:
If I pass "starting_after" into an Stripe::XXXX.all call, it will return objects that are "newer" than the object specified in call. This is not how Stripe has implemented the functionality, after asking their developers, it's poor wording of their API docs, but is working as intended. The docs say the following "The starting_after parameter returns objects created after the named object, in ascending chronological order". What is actually returned is objects returned that are OLDER than the object specified but start AFTER the object in terms of pages as if the pages were organized in descending chronological order. Code example:
```
event1 = StripeMock.mock_webhook_event("customer.created")
event2 = StripeMock.mock_webhook_event("customer.discount.created")
assert StripeEvent.all(starting_after: event2.id).data.count > 0
```

You can test this on stripe to confirm.

PR #407, opened 2017-01-17
Have the data list object behave like Stripe by always putting "creat…

psusmars:
…ed" in descending order

- Otherwise same time events will be assumed that when passed to the list,
  they have been pushed onto the list (where oldest will be first). This is
  to replicate the exact behavior of stripe

See #406


psusmars:
Much like pull request #405, I am receiving errors in the spec related to retrieving coupons that has nothing to do with anything I changed.

alexmamonchik:

psusmars:
Done!

PR #408, opened 2017-01-18
Allow setting order status with a string

psi:
This fixes a typo where the method `includes?` was called instead of `include?`, which isn't valid. Additionally, this allows me to set an order status with a string, rather than a symbol, as I would probably do when working with the API directly. Looking briefly through the Subscriptions request handler, it seems that strings are used for status there as well.


alexmamonchik:

psi:
@alexmamonchik Done

Issue #403, opened 2017-01-06
Invoice.pay should create a new mocked charge


typeoneerror:
Current handler assigns a hard-coded charge ID to the invoice.

https://github.com/rebelidealist/stripe-ruby-mock/blob/master/lib/stripe_mock/request_handlers/invoices.rb#L55

PR #411, opened 2017-02-26
Create a new charge each time an invoice is paid

justin-rhoades:
* addresses issue #403 

* previously, the Invoices#pay_invoice method hard coded a charge id,
this change creates a dynamic charge object each time the method is
called.

* if there is not a customer associated with the invoice, it creates an
anonymous charge by using the generate_card_token helper

* if there is a customer associated with the invoice, it will create a
charge using the customers default payment method, the same behavior
stripe implements


alexmamonchik:
👍

alexmamonchik:

justin-rhoades:
Gotcha, sorry - missed that last commit

typeoneerror:
Great work, @justin-rhoades! Thanks for fixing and closing my issue.

PR #428, opened 2017-04-21
Conversion rate

vkinelev:
#### Summary

Gives app developers an ability to write spec expectations for amounts converted from a charge currency to a destination/account currency.  

#### Details

We faced up with a task to have reports aggregating amounts of charges in different currencies. Obviously, all amounts have to be converted to a destination currency before they can be summed up. Stripe holds amounts in a destination currency in a `Stripe::BalanceTransaction.amount`. It's handy to store converted amount in app's DB for performance reasons.

`stripe-ruby-mock` stores in `Stripe::BalanceTransaction.amount` the same value as an underlying charge amount, so it is not possible to write expectations in specs if a given value is actually taken from the right place. This PR adds an ability to have a precise control over calculation of  `Stripe::BalanceTransaction.amount`: 

```ruby
# New method makes Stripe Mock to calculate `Stripe::BalanceTransaction.amount`
# as `amount = charge_amount * conversion_rate`. It gives app developers a predictable
# way to write spec expectations for converted amounts.  
StripeMock.set_conversion_rate(1.2)

charge = Stripe::Charge.create({
  amount: 
  currency: 'eur'
  source: stripe_helper.generate_card_token
})
bal_trans = Stripe::BalanceTransaction.retrieve(charge.balance_transaction)
# Charge amount and rate are known, thus it is possible to check 
# `Stripe::BalanceTransaction.amount`
expect(bal_trans.amount).to eq(charge.amount * 1.2)
```

### DIFF
--- a/stripe-ruby-mock.gemspec
+++ b/stripe-ruby-mock.gemspec
@@ -24,4 +24,5 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'rspec', '~> 3.1.0'
   gem.add_development_dependency 'rubygems-tasks', '~> 0.2'
   gem.add_development_dependency 'thin', '~> 1.6.4'
+  gem.add_development_dependency 'pry'
 end
--- a/stripe-ruby-mock.gemspec
+++ b/stripe-ruby-mock.gemspec
@@ -24,5 +24,4 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'rspec', '~> 3.1.0'
   gem.add_development_dependency 'rubygems-tasks', '~> 0.2'
   gem.add_development_dependency 'thin', '~> 1.6.4'
-  gem.add_development_dependency 'pry'
 end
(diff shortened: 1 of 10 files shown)


stripe-ruby-mock.gemspec
```diff
@@ -24,4 +24,5 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'rspec', '~> 3.1.0'
   gem.add_development_dependency 'rubygems-tasks', '~> 0.2'
   gem.add_development_dependency 'thin', '~> 1.6.4'
+  gem.add_development_dependency 'pry'
```
alexmamonchik: do we really need this?

vkinelev: Forgot to remove that before pushing. Removed and committed :) Good catch @alexmamonchik !

alexmamonchik: thank you for help!

PR #433, opened 2017-05-18
Stripe 2.8 support

### LINKED ISSUE #410: Stripe 2.0.0 released, updated dependencies
Stripe released version 2.0.0 of the gem recently with the following changes:

* Drop support for Ruby 1.9
* Allow HTTP client that makes Stripe calls to be configured via Faraday
* Drop RestClient
* Switch to OpenAPI 2.0 spec and generated fixtures in test suite
* Switch to Webmock in test suite

`stripe-ruby-mock` has us locked into `~> 1.31`. Any plans to update that dependency? If the 1.9 support is a concern maybe fork the current version of this gem and have it retain 1.9 support while this one moves forward with the `stripe` gem itself?

Borzik:
This PR fixes all references to methods that were removed (or updated) from Stripe v2.0.0 until v2.8.0.
I've tried to avoid any unnecessary changes (except of some trailing spaces that were removed by the editor automatically).
Public API hasn't been changed as well as configuration.

UPD: I'd appreciate if someone who is familiar with the project could help me decide what to do with deprecated recipients. I've disabled some specs because they are no longer passing (you can see the same issue in #423). Fixes #410

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,12 @@ dist: trusty
 sudo: required
 language: ruby
 rvm:
-  - 1.9.3
   - 2.0.0
-  - 2.1.6
+  - 2.1.10
+  - 2.2.7
+  - 2.3.4
 before_install:
-  - rvm 1.9.3 do gem install mime-types -v 2.6.2
+  - rvm 2.1.10 do gem install mime-types -v 2.6.2
   - gem install bundler
 before_script:
   - "sudo touch /var/log/stripe-mock-server.log"
@@ -15,7 +16,7 @@ script: "bundle exec rspec && bundle exec rspec -t live"
 
 env:
   global:
-    - IS_TRAVIS=true STRIPE_TEST_SECRET_KEY_A=sk_test_sXdhUWu3NhrB7r1tkK0zZfMW STRIPE_TEST_SECRET_KEY_B=sk_test_v2K2hV9VJ8AoRKTeDRkQ3rG3 STRIPE_TEST_SECRET_KEY_C=sk_test_QNgKQJhW4fWYT63be2AUSVjJ
+    - IS_TRAVIS=true STRIPE_TEST_SECRET_KEY_A=sk_test_sXdhUWu3NhrB7r1tkK0zZfMW STRIPE_TEST_SECRET_KEY_B=sk_test_uPfIX9ziFNloXwtSdDPJTdnh STRIPE_TEST_SECRET_KEY_C=sk_test_waSy1TaP2RNEpoz9t2pFCysm STRIPE_TEST_SECRET_KEY_D=sk_test_Z1mQZNehRFmI3EN9mHnMafnq
 
 notifications:
   webhooks:
--- a/stripe-ruby-mock.gemspec
+++ b/stripe-ruby-mock.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.require_paths = ['lib']
 
-  gem.add_dependency 'stripe', ['>= 1.31.0', '<= 1.58.0']
+  gem.add_dependency 'stripe', '>= 2.0.3'
   gem.add_dependency 'multi_json', '~> 1.0'
   gem.add_dependency 'dante', '>= 0.2.0'
 
--- a/lib/stripe_mock/api/client.rb
+++ b/lib/stripe_mock/api/client.rb
@@ -8,7 +8,7 @@ module StripeMock
     return false if @state == 'live'
     return @client unless @client.nil?
 
-    alias_stripe_method :request, StripeMock.method(:redirect_to_mock_server)
+    alias_stripe_method :execute_request, StripeMock.method(:redirect_to_mock_server)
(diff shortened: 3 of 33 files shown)


benoror:
@gilbert @alexmamonchik any plans to ship this soon? We need to upgrade to latests stripe gem version, and this is a blocker for us. Any info would be appreciated, regards!

nlively:
@Borzik, I tried using this branch to mock Stripe's SubscriptionItem API, and had some issues.  Code is here: https://github.com/boomr/stripe-ruby-mock.  I have 1 additional commit since this PR, but I don't think it's merge-worthy.  Would appreciate your insights on this to make sure i'm following existing patterns with this gem and to help round out on SubscriptionItems.

.travis.yml
```diff
@@ -1,10 +1,10 @@
 language: ruby
 rvm:
-  - 1.9.3
-  - 2.0.0
-  - 2.1.6
+  - 2.1.10
```
alexmamonchik: actually, I think we still need to support ruby 2.0.

stripe-ruby-mock.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.require_paths = ['lib']
 
-  gem.add_dependency 'stripe', '~> 1.31'
+  gem.add_dependency 'stripe', '~> 2.8'
```
alexmamonchik: also, I think minimum required stripe gem version is 2.0.0, not 2.8.

.travis.yml
Borzik: Yes, I've been thinking of adding ruby 2.0.x there too, but since we've got only three api keys, I've decided to go with ruby 2.1 and higher. Could you please give me a test key to use with ruby 2.0.x? We also have 2.4 already, so probably we'll need to have it as well; then I need two more test api keys.

alexmamonchik: try those keys, but I'm not sure https://github.com/rebelidealist/stripe-ruby-mock/blob/updated-stripe-keys/.travis.yml#L16

Borzik: Yes, I already use them for rubies 2.1, 2.2 and 2.3. If it's ok to use them to test against two ruby versions simultaneously, I can do it, although I'm not sure it won't create any overlaps. I've taken these from your PR to have the latest Stripe API version for testing.

alexmamonchik: Actually, we have key for ruby 2.0 - you just removed it :-) https://github.com/rebelidealist/stripe-ruby-mock/blob/master/.travis.yml#L16

Borzik: yes, I know :) I just don't know which Stripe API version is attached to those keys. E.g. Recipients feature may work on some older APIs, but will not work now, so I want to make sure it will test against an up-to-date API version.

stripe-ruby-mock.gemspec
Borzik: I've ran test suite against `2.0`, but it had some specs failed, and it worked fine with `2.1` only. I guess there were some fixes in 2.1. 
There's not such a big difference between 2.0 and 2.1, so I think it will worth upgrading to 2.1 or even to the most recent version, but 2.1 should work since tests don't fail.

alexmamonchik: existing code works fine with stripe 1.58. With 2.0 it totally broken. The main problem in ```ruby
Stripe::StripeClient.active_client.method(:request)
```
you fixed it. I want to make a release with 1.31-1.58 stripe gem for old projects and 2.0 and newest for new projects. You set 2.1 as minimum and it means that people who use 2.0 stripe can't use any of stripe-ruby-mock release.

Comment (alexmamonchik) [inline stripe-ruby-mock.gemspec]: what's wrong with 2.0? Let's fix it and merge this PR

Comment (Borzik) [inline stripe-ruby-mock.gemspec]:
``` 
=== 2.0.3 2017-03-16
* Fix marshalling of `StripeObjects` that have an embedded client
```
I've tested it with 2.0.3 and tests didn't fail, but failed with 2.0.2. So I think that's the issue that causes some specs to fail. I've lowered required version to 2.0.3

alexmamonchik: perfect! Thanks

alexmamonchik:
@Borzik please, pull master - it has some travis updates

alexmamonchik:
@Borzil we are on the last step before merge. Please, pull master one more time

Borzik:
@alexmamonchik 
Done

mcmire:
@Borzik Just wanted to clarify - if I understand correctly, this PR fixes any incompatibilities/deprecations with Stripe 2.8 but doesn't necessary add anything that versions 2.0 through 2.8 have added?

alexmamonchik:
@mcmire also, updated stripe keys for live tests, locked minimum stripe gem requirements

Borzik:
@mcmire yes, this was intended to fix incompatibilities with stripe versions 2.0.0 to 2.8, it doesn't add any new features.

alexmamonchik:
follow #442 PR. from this we'll use newest API and stripe versions

mcmire:
Awesome sauce. Thanks for doing this by the way!

PR #435, opened 2017-05-30
added test for negative amount of transfer

alexmamonchik:
- test for case when transfer uses negative amount


spec/shared_stripe_examples/transfer_examples.rb
```diff
@@ -78,4 +78,20 @@
       expect(e.http_status).to eq(400)
     }
   end
+
+  it 'when amount is negative', focus: true, live: true do
```
Borzik: @alexmamonchik 
This is the second place which has `focus: true`.

PR #436, opened 2017-05-30
raise error when attempting to create a new subscription with wrong

### LINKED ISSUE #413: Doesn't break on invalid params when creating a subscription
Using the latest gem and latest Stripe API

Mock should break when attempting to create a new subscription with the params `trail`

The correct params is `trial_in_days`

alexmamonchik:
fixed #413

### DIFF
--- a/spec/shared_stripe_examples/subscription_examples.rb
+++ b/spec/shared_stripe_examples/subscription_examples.rb
@@ -215,6 +215,18 @@ shared_examples 'Customer Subscriptions' do
       }
     end
 
+    it 'when attempting to create a new subscription with the params trial', focus: true, live: true do
+      plan = stripe_helper.create_plan(id: 'trial', amount: 999)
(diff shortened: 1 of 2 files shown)


spec/shared_stripe_examples/subscription_examples.rb
```diff
@@ -215,6 +215,18 @@ def gen_card_tk
       }
     end
 
+    it 'when attempting to create a new subscription with the params trial', focus: true, live: true do
```
Borzik: @alexmamonchik by `focus: true` you've disabled 90% of all tests in Travis builds (which will fail if you remove it), only 60 tests are being executed now. Where's one more place in specs which has `focus: true`.

alexmamonchik: thank you. Stupid mistake. fixed

PR #449, opened 2017-07-12
Transfer response should include a metadata field.

dorkrawk:
Transfer response now includes a `metadata` field, as specified by the Stripe docs: https://stripe.com/docs/api#transfer_object


dorkrawk:
It looks like the build failure came from different Ruby versions not being installed on the CI server. Does that seem right?

alexmamonchik:
temporary fixed, please, pull master. Opened issue #450

dorkrawk:
@alexmamonchik ok, the tests look good now.

PR #457, opened 2017-08-01
When setting a nil tos date stripe returns a 400 error

arthurnn:
This simulates that same behaviour, so we can catch errors like that in tests.
I copied the same error message and status. The response from Stripe is:

```
{:error=>{:type=>"invalid_request_error", :message=>"Invalid integer: ", :param=>"tos_acceptance[date]"}}
```


arthurnn:
any response on this?

thanks ❤️

PR #460, opened 2017-08-09
WIP: Automatically mark past due subscriptions.

dblock:
Trying to implement something that automatically marks subscriptions past due. I think there's a better place to put this - where can I plug a helper that examines any subscription data being returned and potentially alters it? 

In some ways this is what Stripe would do - you create a subscription, time goes by, subscription expires.


alexmamonchik:
hmmmm, look at  [this](https://github.com/dblock/stripe-ruby-mock/blob/past-due-subscriptions/lib/stripe_mock/request_handlers/subscriptions.rb#L101) 
```ruby
subscription.merge!(custom_subscription_params(plan, customer, params))
```
we use `custom_subscription_params` and 
```ruby
start_time = options[:current_period_start] || Time.now.utc.to_i
```
there is the problem - Timecop doesn't affect on that and when you check:
```ruby
if Time.at(subscription[:current_period_end]) < Time.now.utc
```
`Time.now.utc` still the same time as before Timecop call.

dblock:
I looked into what you suggested, but that happens when we make up a subscription. I think we need to stick to how Stripe works, ie. at some point in time you want to retrieve a subscription that was previously `active` and now it's `past_due` or `canceled`.

I have updated the PR to reflect that, LMK what you think and if there's a way to do this cleaner?

PR #471, opened 2017-09-10
Make "created" field on subscriptions dynamic

Borzik:
Sometimes we may need to set subscription's `created` value to some specific date/time (e.g. if we want to figure out if a subscription is newly created or has already been renewed).

### Before:
Subscription's `created` field always had constant value which was impossible to change.

### After:
You can provide `created` in `Stripe::Subscription.create` call. By default, it will be set to current time and will be equal to `current_period_start`.

### DIFF
--- a/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
+++ b/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
@@ -9,8 +9,10 @@ module StripeMock
       def custom_subscription_params(plan, cus, options = {})
         verify_trial_end(options[:trial_end]) if options[:trial_end]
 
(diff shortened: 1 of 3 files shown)


alexmamonchik:
hmm, so specific case. Test, please.

Borzik:
It works exactly the same as `current_period_start` subscription attribute for which I haven't been able to find any tests. 
I think `created` attribute can be expected to be set to creation time, not to some specific date in the past, so I'm not sure it's so specific case.

alexmamonchik:
ok, let's add that but can you add confirmation test for your code?

Borzik:
@alexmamonchik could you please check it now?

alexmamonchik:
perfect! thanks!

PR #473, opened 2017-09-17
Update Stripe::Transfer with latest API attributes

abevoelker:
I noticed transfer attributes were out of date when I tried to access `created` but that didn't exist: https://stripe.com/docs/api#transfer_object

I left `recipient` in since that broke some tests and I'm not familiar enough with the code to do a better refactoring. Sorry if this is incomplete or insufficient with how the code handles API versioning or something. But it got me what I needed for mocking.

### DIFF
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -592,31 +592,29 @@ module StripeMock
       currency = params[:currency] || 'usd'
       id = params[:id] || 'tr_test_transfer'
       {
-        :status => 'pending',
         :amount => 100,
-        :account => {
-          :object => 'bank_account',
-          :country => 'US',
-          :bank_name => 'STRIPE TEST BANK',
-          :last4 => '6789'
-        },
-        :recipient => 'test_recipient',
-        :fee => 0,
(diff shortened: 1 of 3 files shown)


alexmamonchik:
thank you. Can you add some tests for your changes. Please, look at this doc https://github.com/rebelidealist/stripe-ruby-mock/wiki/Implementing-a-New-Behavior

abevoelker:
Okay updated with some tests. I ended up removing `recipient` after all and getting the skipped tests working using the `destination` parameter instead.

When I run the live tests I get some failures for the "Invoice API" but those seem to be currently failing on master so I think unrelated? All the transfer stuff seems to pass anyway.

abevoelker:
Pushed one more change to fix the CI error. Was testing locally on 2.4 and didn't think about older Ruby versions not having the string key syntax

abevoelker:
Current CI failure seems to be an issue with the CI server's Stripe account (works fine for mine):

```
Stripe::InvalidRequestError: You can only create new accounts if you've registered your platform, which you can do at https://dashboard.stripe.com/account/applications/settings.
```

PR #475, opened 2017-09-29
Don't store upcoming invoice in memory

cgunther:
The invoice does not actually exist, but is just a preview. It's not an invoice that should be retrieved along with all other invoices for a given customer.

Previously, if you tried to list all invoices for a customer after fetching the upcoming invoice, the upcoming invoice would be included in the list of all invoices for that customer.

For now I left the invoice ID in place, however in my testing, Stripe does not return an ID for an upcoming invoice (likely because it can't be retrieved by ID), however their API documentation suggests the return JSON of fetching the upcoming invoice would have an ID: https://stripe.com/docs/api#upcoming_invoice

I had to change fetching line items for an upcoming invoice so that it'd use the line items in the response, not query the API to get the line items, since that'd fail the test as you can't fetch line items for an invoice that is not stored in memory.

### DIFF
--- a/lib/stripe_mock/request_handlers/invoices.rb
+++ b/lib/stripe_mock/request_handlers/invoices.rb
@@ -73,7 +73,7 @@ module StripeMock
           else
             customer[:subscriptions][:data].min_by { |sub| sub[:current_period_end] }
           end
-        
+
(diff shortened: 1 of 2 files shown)


cgunther:
Thanks for the quick merge!

PR #476, opened 2017-09-29
Support previewing invoice while prorating new plan with same interval

cgunther:
When previewing a change to a new plan with the same interval, a line item is added for the remaining time in the period based on the new plan's amount, so long as there is no trial.


cgunther:
Thanks for the quick merge!

PR #477, opened 2017-10-01
Mock refund reason

PlugIN73:
To issue #421


alexmamonchik:
thank you. Please, add a confirmation test for your changes

PlugIN73:
@alexmamonchik, hi! I'm racking my brain but can't come up with confirmation test. This method ```mock_charge``` is called in many test methods and we can guarantee syntax by passed tests (I checked this). So I think there is no proof needed. If you think otherwise, can you advise me?

arthurnn:
👍

PR #491, opened 2017-10-30
Add balance transaction to Transfer object

arthurnn:


alexmamonchik:
`undefined method `abs' for "100":String
`

arthurnn:
I will fix it.

arthurnn:
OK, fixed.. The issue was that we validated the amount had to be a Integer or to be able to transform it to an Integer, but I was not calling `to_i` before sending to the balance transaction instantiation method.

alexmamonchik:
great. Can you add a test for your changes?

PR #493, opened 2017-11-28
Add outcome field to Charge

jbourassa:
The field is documented in the API doc:
https://stripe.com/docs/api#charge_object-outcome

-------

Not sure why the tests are failing – I don't think the `outcome` param should ever be sent when saving, it is readonly. I'd be happy to work on fixing the tests for this, but I will need a pointer before; I dug quite a bit and couldn't figure out.


jbourassa:
Not going to work on this anytime soon, I'll just close this PR.

PR #494, opened 2017-12-03
Add basic no-op endpoints for subscription items

lfittl:
This adds enough support to make some tests pass that I have in a test suite:

`get /v1/subscription_items` (returning the `@subscription_items` from the mock instance)
`post /v1/subscription_items` (doing nothing and returning empty `{}`)

Its nowhere near where it needs to be, but seems like an easy enough start.


alexmamonchik:
have you passed tests locally?

alexmamonchik:
added tests

PR #505, opened 2018-01-17
Support for items in subscription create/update

justin808:
Following change allows items to provide the plan.


justin808:
@alexmamonchik I added tests.

PR #509, opened 2018-01-29
Remove unused changelog

pwim:
While it would be nice to have a changelog, since this file has never been touched since the initial release, it creates confusion.


alexmamonchik:
I agree with it. Most of changes are pull requests and a little bit hard to follow all changes in release.

PR #512, opened 2018-02-05
Webhook support for API 2018-01-23

archonic:
I was running into some problems on API 2018-01-23 with the differences between mocked and real webhooks, so I went through and copied all the example requests.


archonic:
Closing until I handle tests.

PR #513, opened 2018-02-06
Add Stripe::Account #delete endpoint

kssmilik:
adds Stripe::Account #delete endpoint


alexmamonchik:
thank you. Please, add a test for your changes.

PR #515, opened 2018-02-10
Support for API change for 2018-02-06

justin808:
* Stripe.api_version = "2018-02-06"
* Place plan name in a hash under key product when
creating a plan.
* Plan objects come back with a product id rather
than a name.

* Per Stripe Docs:

Changes since API version 2017-12-14:

Each plan object is now linked to a product object with type=service.
The plan object fields statement_descriptor and name attributes have been
moved to product objects. Creating a plan now requires passing a product
attribute to POST /v1/plans. This may be either an existing product ID
or a dictionary of product fields, so that you may continue to create
plans without separately creating products.

### DIFF
--- a/lib/stripe_mock/test_strategies/base.rb
+++ b/lib/stripe_mock/test_strategies/base.rb
@@ -6,7 +6,9 @@ module StripeMock
         currency = params[:currency] || 'usd'
         {
           :id => 'stripe_mock_default_plan_id',
-          :name => 'StripeMock Default Plan ID',
+          :product => {
+            :name => 'StripeMock Default Plan ID'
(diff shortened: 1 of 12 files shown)


alexmamonchik:
Thank you for help. But you added breaking changes. We can release it in new major version but please, adjust other tests for new `product`  attribute

sbounmy:
👍 running Stripe ```2018-02-06``` having the same problem with plan / product

ac21:
+1; running Stripe `2018-02-28`; having the same problem with plan / product.  will try to build on this commit request to adjust other tests for new `product` attribute

Issue #516, opened 2018-02-15
StripeMock and Rails System Tests


johnmaxwell:
First of all, thank you for stripe-ruby-mock. It has saved me a lot of time and led to more complete test coverage across several projects that use Stripe.

I'm having difficulty getting StripeMock working with the Capybara-based testing that underpins Rails system tests. There are several pieces at play that are contributing to the issue, and I recognize that this difficulty doesn't lie completely with stripe-ruby-mock. I'm primarily posting this issue with the hopes of finding some guidance to resolving this issue.  Here are the versions that I'm using.

* stripe-ruby-mock (2.5.2)
* stripe (3.9.2)
* rails (5.1.4)

The core of the issue seems to boil down to this:

1. Stripe implements per-thread Stripe::StripeClient.active_client as of https://github.com/stripe/stripe-ruby/pull/591
2. The test server started by Rails System Tests run in a separate thread pool under Puma.
3. StripeMock.start and StripeMock.start_client intercept calls Stripe::StripeClient.active_client of the current thread but not the threads of the spawned test server. 

So even if I start StripeMock with StripeMock.start or StripeMock.start_client in the test file, the puma-based test server running in another thread doesn't have same StripeClient with the intercepted `execute_request` method. 

At present, it looks like the StripeClient is stored in either the stripe_client or stripe_client_default_client fiber-local variables. Would the idea be to start enable StripeMock in the integration test file and then overwrite one of those fiber-local variables in the puma-based test server? If so, any guidance for how to do this? Or am I off-base here?

nfm:
I don't have too much to add, but I also ran into this issue today and can confirm you're on the right track @johnmaxwell. I've downgraded to stripe 3.5.0 as a workaround for now.

If you haven't run into this already, be aware that if you're using any of Stripe's JS libraries to handle cards they're also difficult to mock. They'll send the card details to Stripe directly and return a real token, which stripe-ruby-mock then needs to be able to deal with. More info in https://github.com/rebelidealist/stripe-ruby-mock/issues/360.

cgunther:
I believe I have a fix for this in #518. Could you test with my branch and see if that resolves the issue for you?

nfm:
[...]

Issue #508, opened 2018-01-27
No such plan: stripe_mock_default_plan_id when running in system Specs


Bodacious:
When I try to test my registration process using RSpec system specs, I keep getting the same error: 

```
Stripe::InvalidRequestError:
            No such plan: [plan id]
```

### My spec helper:

```
require 'thin'
require 'stripe_mock'
require_relative "stripe_helpers"

StripeMock.spawn_server

RSpec.configure do |config|
  config.before(:each, type: :system) do
    StripeMock.start_client
  end
  config.after(:each, type: :system) do
    StripeMock.stop_client
  end
  config.include(StripeHelpers, type: :system)
end
```

### My system test:

```
background do
  spec_helper = StripeMock.create_test_helper
  plan = spec_helper.create_plan({
      id: SecureRandom.hex,
      amount: 945,
      interval: :month,
      interval_count: 1,
      livemode: false,
      trial_period_days:30,
    })
    ENV["STRIPE_INITIAL_PLAN_ID"] = ensure_plan.id
end

scenario "artist creates a new account", :js do
  # simulate signing up
end
```

### My app code:

```
class RegistrationJob < ActiveJob

INITIAL_PLAN_ID = ENV["STRIPE_INITIAL_PLAN_ID"]
TRIAL_PERIOD    = 1.month.freeze

class MissingStripeRecordError < StandardError
end

queue_as :default

attr_reader :account

attr_reader :user

attr_reader :customer

attr_reader :subscription

attr_reader :plan_id


def perform(account, plan_id: INITIAL_PLAN_ID, trial: false)
  @account = account
  @user    = account.owner
  @plan_id = plan_id
  @trial   = trial
  unless account.stripe_customer_id.to_s == stripe_customer.id.to_s
    logger.debug("Updating customer Stripe ID")
    account.update(stripe_customer_id: stripe_customer.id)
  end
  account.subscriptions.order('created_at DESC')
                   .first_or_create!(
                     stripe_subscription_id: stripe_subscription.id,
                     stripe_plan_id:   stripe_plan.id,
                     trial_expires_at: Time.at(stripe_subscription.trial_end.to_i),
                     expires_at:       Time.at(stripe_subscription.current_period_end)
                    )
end

def trial?
  @trial
end


private


  def stripe_customer
    @stripe_customer ||= begin
      if @account.stripe_customer_id?
        return_val = Stripe::Customer.retrieve(@account.stripe_customer_id)
        unless return_val
[...]

PR #518, opened 2018-02-17
Support mocking Stripe requests in separate threads

### LINKED ISSUE #504: Using stripe-ruby v3.5.1: Not responding when called inline with Sidekiq
v3.5.1 fixes an issue with [thread safety](https://github.com/stripe/stripe-ruby/issues/590). For some reason this results in no response when calling Stripe commands after upgrading to this version (or newer). It just stops (doesn't execute beyond Stripe calls). I am only able to replicate this issue when running Stripe commands from integration tests with code in a Sidekiq job which I am running as inline (Sidekiq::Testing.inline!).

cgunther:
As of stripe v3.5.1 (https://github.com/stripe/stripe-ruby/commit/a210c5cd76f9178b5253b01c845076f2ac3f5e0a), the default client was memoized per-thread, therefore we were only mocking requests made on the same thread the tests ran on. If a Stripe request was made in a separate thread (such as a Rails feature/system test), it'd go right to Stripe.

Now we'll mock requests from any Stripe client instance instead of just the one for this thread.

I'm not sure if this is the right/best way to do this, using the private `define_method` seems wrong, but this was enough to solve the problem.

Probably fixes #504, #508, and #516.

### DIFF
--- a/lib/stripe_mock/api/client.rb
+++ b/lib/stripe_mock/api/client.rb
@@ -8,7 +8,7 @@ module StripeMock
     return false if @state == 'live'
     return @client unless @client.nil?
 
-    alias_stripe_method :execute_request, StripeMock.method(:redirect_to_mock_server)
+    Stripe::StripeClient.send(:define_method, :execute_request) { |*args| StripeMock.redirect_to_mock_server(*args) }
     @client = StripeMock::Client.new(port)
     @state = 'remote'
     @client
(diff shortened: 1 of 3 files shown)


johnmaxwell:
I can confirm that this fixes the issue I reported in #516 . Thanks!

cgunther:
Thanks for the quick merge!

PR #523, opened 2018-02-26
Sensible error message when plan's currency doesn't match user's urrency

pwim:
Having my tests fail with `(Status 400) lol` was confusing for me.


Bodacious:
+1

amrdruid:
I was going to open a PR for this as well, it's confusing indeed. Thanks for making it more clear

PR #529, opened 2018-03-16
Changes BalanceTransaction source from card object to charge id

alienfungi:
It seems like the BalanceTransaction source should be the charge id, not the card object.


alexmamonchik:
please, update related tests

alienfungi:
Sorry, I didn't notice that failing spec due to the noise from lots of the tests failing before I even changed anything.  I was receiving a too many files open error from running your test suite.  Had to raise my open file limit from 1024 to ~2000 to get tests to pass.  I fixed the spec that I caused to fail and amended my commit.

PR #533, opened 2018-04-17
Support billing_cycle_anchor on subscriptions

cgunther:
This allows indicating when the first full payment should be charged.

It was documented starting in API version [2018-02-05](https://stripe.com/docs/upgrades#2018-02-05), but is available retroactively on past versions.

When set, the current period ends at the given time, before a new, full period starts.

Technically it can be combined with trials, but I'm not as familiar with that side, so I didn't attempt to implement it for fear of implementing it wrong.


cgunther:
Thanks for the quick merge!

PR #560, opened 2018-10-01
added dynamically generated exp_year

shohey1226:
My test was failed after Oct 2018 is reached. Updated to fix this.


shohey1226:
just found this is dup. closing mine.

PR #587, opened 2019-03-26
Subscription Creation with multiple plan identifiers

aliahmed922:
This PR fixes the `TypeError (no implicit conversion of Symbol into Integer)` error, raised when adding multiple plan identifiers in `Stripe::Subscription.create` method.

```ruby
Stripe::Subscription.create(customer: customer.id, items: [{ plan: 'my_plan'}])
# => TypeError (no implicit conversion of Symbol into Integer)
```

### DIFF
--- a/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
+++ b/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
@@ -10,6 +10,7 @@ module StripeMock
         subscription.merge!(custom_subscription_params(plans, customer, options))
         subscription[:items][:data] = plans.map do |plan|
           if options[:items] && options[:items].size == plans.size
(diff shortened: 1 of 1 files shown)


lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
```diff
@@ -10,6 +10,7 @@ def resolve_subscription_changes(subscription, plans, customer, options = {})
         subscription.merge!(custom_subscription_params(plans, customer, options))
         subscription[:items][:data] = plans.map do |plan|
           if options[:items] && options[:items].size == plans.size
+            options[:items] = options[:items].values if options[:items].respond_to?(:values)
```
aliahmed922: `options[:items]` is returning a Hash(`{"0": { plan: 'my_plan' }}`).

PR #557, opened 2018-09-27
Products and Plans

s2t2:
Additions:

  + Enables mocking of`Stripe::Product` objects.

Revisions:

  + Requires an updated version of the Stripe ruby gem.
  + Updates `Stripe::Plan` expectations to validate associated product.

PR #593, opened 2019-05-10
2019-03-14 API support (incl. Products)

Borzik:
I used #557 as a starting point, fixed all specs and updated all API keys to the newest (2019-03-14) API version.
Notice: this is a large update and it needs additional testing before it can be merged to master. I'm not sure if I will be able to test it all myself (especially parts I don't personally use in my projects, e.g. upcoming invoices).
If anyone uses this branch in their project, feel free to share your feedback.

I have also changed Travis config to test against current Ruby versions only (e.g. 2.3 is outdated already).

I'm also thinking of new versioning system. Raw SemVer doesn't work very well because we mostly depend on Stripe API version identifier, which looks like date. So upgrading from `2.4.0` to `2.5.0` doesn't make much sense unless you know which Stripe API version corresponds to `2.4.0` and `2.5.0`, so you can only check changelog of both Stripe API and stripe-ruby-mock. Instead, we can include current Stripe API version into our version number, e.g. have `2.7.0-190314` for this release (or add some tags/table explaining versioning), so it would be clear that people using 2018-XX-XX API version will most likely experience issues; and people using 2019-10-10 version in the future will probably have to update some parts of stripe-ruby-mock to mock this version of API; it will also be expected that pull requests will have all tests passing on given Stripe API version, like this PR which fixes all tests for 2019-03-14. But this is a huge update and I'm not sure if there will be a collaborator who is ready to take this project and manage it anytime soon.

### I think it needs a bit more work/testing (including major version bump), so don't merge it for now please. If anyone could use this fork to test on their own project and give feedback, I'd really appreciate that.


alexmamonchik:
@Borzik I'm definitely agree  with you about versioning system.

PR #536, opened 2018-04-28
Support Product for API change for 2018-02-06

sbounmy:

PR #595, opened 2019-05-15
Support for Products and SKUs

evanmightybell:
This pull request adds support for products and skus.

The product code was written by @sbounmy and the original pull request for that is https://github.com/rebelidealist/stripe-ruby-mock/pull/536

SKUs are dependent on products, so it makes sense to bundle them into the same pull request. Here's the Stripe documentation for SKUs: https://stripe.com/docs/api/skus


alexmamonchik:
@evanmightybell ooops. Looks like I've merged newest PR #601 and it has conflicts with your. Sorry, please correct but remember that this gem supports api version [2017-06-05](https://github.com/rebelidealist/stripe-ruby-mock#specifications). Let's update the gem api version in next release but keep #601 for current major version. thanks

PR #614, opened 2019-06-27
Support for latest API version

Borzik:
All tests (mock and live) are passing when tested against 2019-05-16. Also:

- added product webhooks
- drop support for `subscription.delete(at_period_end: true)`
- add support for `subscription.cancel_at_period_end = bool`
- updating subscription which has `cancel_at_period_end: true` will not resume it unless `cancel_at_period_end` is explicitly set to `false`
- add product validations
- allow `enable_incomplete_payments` param. It does not affect anything at the moment, but subscription should require card when it's `false`, and should always succeed (with incomplete status) when it's `true` (that is new default)

Also it looks like running two spec suites using same API key concurrently will raise conflicts and tests will fail, so we'll probably need to limit build concurrency on Travis.


joshm1204:
Is PR going o be merged in?  It would be great to have it included now.

Borzik:
@alexmamonchik Is there anything you can do to this PR? 
Maybe create a `2019-05-16` (that's the version API keys added by this PR use) or simply `recent-api` branch and merge it there. And then PRs which add features not supported by on old API will go to the same branch.
Because I've rebased it a few times already, and if #638, #645, #633, or some newer PR will be merged, new conflicts will be introduced and I'll probably just keep using my fork because rebasing this thing takes quite a lot of time.

gravitystorm:
I didn't notice this PR when I created #645 but it seems we are in agreement about updating the tested ruby versions! :smile:

It would be great to have feedback from the maintainer as to his preferences for situations like this, whether one big PR is preferred, or have it broken into smaller pieces for different topics. I think we also have a bit of duplication between this and other pending pull requests, but I'm not sure what we can do about that. But feedback from @alexmamonchik is probably the most important thing.

alexmamonchik:
Thank you, guys for that. We discussed with @PlugIN73 how to follow recent Stripe api and want to build a new major version `3.0.0` with newest api support. I suggest to merge this PR to version 3. What do you think about that?

gravitystorm:
Sounds good to me!

Borzik:
Sounds great, thanks for response! Please let me know if there's anything I have to do.

alexmamonchik:
thank you guys for this PR. Merged! I've just fixed conflicts.

Issue #618, opened 2019-07-11
Payment Intents and Methods APIs (SCA)


Jensenks:
Is anyone working on the new SCA ready APIs? 

**Edit:** Looks like it's getting worked on by contributors here: https://github.com/FarmDrop/stripe-ruby-mock/commits/payment_method

It would be great to have these added to stripe-ruby-mock before the [Strong Customer Authentication](https://stripe.com/docs/strong-customer-authentication) regulation comes into effect on September 14, 2019.

The APIs include:
[Stripe Payment Intents API](https://stripe.com/docs/payments/payment-intents)
[Stripe Payment Methods API](https://stripe.com/docs/payments/payment-methods)

ashkan18:
Opened a PR to add support for it 🤞

ashkan18:
this adds support for more SCA use cases https://github.com/rebelidealist/stripe-ruby-mock/pull/622

PR #619, opened 2019-07-15
Add support for PaymentIntents

ashkan18:
# Change
Add Basic Support for [`PaymentIntents` API](https://stripe.com/docs/api/payment_intents) following [contribution doc](https://github.com/rebelidealist/stripe-ruby-mock/wiki/Implementing-a-New-Behavior#contributing-to-stripemock)

Fixes https://github.com/rebelidealist/stripe-ruby-mock/issues/618

# Different Statuses
As part of test you may want to get payment intents in `succeeded`, `requires_action` or `requires_payment_method` states. In this current approach when creating a `PaymentIntent` by default you'll get `succeeded` payment intent. 
- setting amount to `3184` (which comes from last 4 digits of Stripe's test card for SCA in [here](https://stripe.com/docs/testing#regulatory-cards)) will set the state to `requires_action`
- setting amount to `3178` (coming from same link above) will pass authentication but fail payment method and set the status to `requires_payment_method`. In this case we also set `last_payment_error` on `PaymentIntent` to a simple `card_declined` error.


ashkan18:
Looks like failures are not related to my changes and happen only on _live_ tests, right?

sofianeOuafir:
what version include your changes @ashkan18 ?

ashkan18:
@sofianeOuafir i think it got released as part of `2.5.8`

sofianeOuafir:
@ashkan18 I have the 2.5.8 but your changes don't seem to be a part of it. For instance, the mock_payment_intent method is not in the stripe-ruby-mock/lib/stripe_mock/data.rb file

sofianeOuafir:
@ashkan18 btw, thanks for you work :) Do you know how can I solve my problem?

ashkan18:
sure! @sofianeOuafir what do you have in your `Gemfile`? we have switched to that and things are working fine for us.

olimart:
@ashkan18 
Thanks for your work. I'm running into the situation where when we confirm the payment intent, Stripe attaches any successful charges to the `PaymentIntent` object.
We rely on status (which becomes `succeeded`) but would also like to make sure charge is included.
Would be great to include that case as well, don't you think?

PR #622, opened 2019-07-18
Improve PayementIntents responses

ashkan18:
# Changes
This is a follow on #619  to do better support for different use cases.

A `PaymentIntent` can have these following statuses:
- `succeeded`: when it was successfully confirmed and captured
- `requires_action`: when it has to go trough SCA process
- `requires_payment_method`: when it was confirmed successfully (doesn't need SCA) but actual processing of charge failed.
- `requires_capture`: when it was confirmed but it was created with `capture_method: manual` so it needs to be _captured_ separately.

In #619 , there was support for `requires_action` but this PR adds support for two other cases.

When using `FAILED_TRANSACTION_AMOUNT` (which is 3178):
- on creating payment intent, it will be confirmed but will fail and have failure reason in `payment_intent.last_payment_error`
- on _capture_, it will go to `requires_payment_method` and have failure reason in `last_payment_error`.

### DIFF
--- a/lib/stripe_mock/request_handlers/payment_intents.rb
+++ b/lib/stripe_mock/request_handlers/payment_intents.rb
@@ -1,6 +1,10 @@
 module StripeMock
   module RequestHandlers
     module PaymentIntents
+      FAILED_TRANSACTION_AMOUNT = 3178
+      REQUIRES_ACTION_AMOUNT = 3184
+      REQUIRES_CAPTURE_AMOUNT = 3169
+
       ALLOWED_PARAMS = [:description, :metadata, :receipt_email, :shipping, :destination, :payment_method, :payment_method_types, :setup_future_usage, :transfer_data, :amount, :currency]
 
       def PaymentIntents.included(klass)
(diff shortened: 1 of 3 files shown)


lib/stripe_mock/request_handlers/payment_intents.rb
```diff
@@ -18,17 +22,19 @@ def new_payment_intent(route, method_url, params, headers)
...
         end
-        last_payment_error = params[:amount] == 3178 ? last_payment_error_generator(code: 'card_declined', decline_code: 'insufficient_funds', message: 'Not enough funds.') : nil
+        last_payment_error = params[:amount] == FAILED_TRANSACTION_AMOUNT ? last_payment_error_generator(code: 'card_declined', decline_code: 'insufficient_funds', message: 'Not enough funds.') : nil
         payment_intents[id] = Data.mock_payment_intent(
           params.merge(
             id: id,
             status: status,
-            last_payment_error: last_payment_error
+            last_payment_error: last_payment_error,
+            charges: status == 'succeeded' ? charges_date : empty_charges
```
Haerezis: You should also add charges in `confirm_payment_intent`

ashkan18: good point!  Thanks! I'll update.

Haerezis: Nice, thanks for the quick update !

PR #624, opened 2019-07-21
Add invoice settings to customers

andreaslillebo:
Customers were missing `invoice_prefix` and `invoice_settings`.

These changes enables a `PaymentMethod` to be set as the default for invoice or subscription payments, by setting `invoice_settings.default_payment_method` on the `Customer`, to the `PaymentMethod`’s ID, as described here:
https://stripe.com/docs/api/payment_methods/attach

### DIFF
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -120,6 +120,7 @@ module StripeMock
     def self.mock_customer(sources, params)
       cus_id = params[:id] || "test_cus_default"
       currency = params[:currency] || StripeMock.default_currency
+      invoice_prefix = SecureRandom.hex(4).upcase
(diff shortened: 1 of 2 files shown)


Haerezis:
Good PR. I had to monkeypatch StripeMock::Data to make my tests pass but I would rather not having to do that :s

renchap:
Could you change this as well?
https://github.com/rebelidealist/stripe-ruby-mock/blob/697f8c0ae1f485c10a4d3af9b7599a144f9d91f6/lib/stripe_mock/request_handlers/subscriptions.rb#L258-L264

Currently if the customer has a default payment method but no default source (ie, only using the new SCA-compatible API), creating a subscription will fail.

andreaslillebo:
Fixed :)

PR #627, opened 2019-07-24
Payment Methods support

rdptz:
- adds essential payment methods supports (create, attach to customer, list)
- fills the payment intent charge object so that the stripe transaction fee can be read using the  Stripe::BalanceTransaction.retrieve(transaction_id) method call.


alexmamonchik:
@rdptz please, review this PR and solve conflicts.

PR #632, opened 2019-07-28
Support Tax Rates in subscriptions

renchap:
Following up on #604, this PR adds tax-related fields on subscriptions & items.


pwim:
It looks like this PR is doing the same thing as #636 (in possibly a more complete way).

PR #636, opened 2019-08-05
Add support for default_tax_rates param when creating subscriptions

lefterisnik:


pwim:
I verified this branch allows me to set the default_tax_rates attribute when creating a subscription, and all of my application's specs pass.

PR #640, opened 2019-08-15
Support Payment Intent status requires_capture

theodorton:
In order to test separate auth and capture, one must be able to create payment intents with the status `requires_capture`.

Based on test card with 4 last 3055:

> 3D Secure authentication may still be performed, but is not required. By default, your Radar rules will not request 3D Secure authentication for this card.
– https://stripe.com/docs/testing#three-ds-cards


theodorton:
The specs fail on my machine as well when running `bundle rspec -t live` both on `master` and my branch.

alexmamonchik:
@theodorton please, pull master

theodorton:
Sorry, this is now rebased on master.

PR #643, opened 2019-08-23
Support stripe-ruby 5.0+ Stripe::CardError.new arguments

j15e:
Updating the mock error system to support stripe-ruby 5+ which was released a few days ago.

@see https://github.com/stripe/stripe-ruby/pull/816 for the specific changes

I made it backward-compatible but I don't know what is you policy about it. Otherwise maybe you should bump a major version of `stripe-ruby-mock`.

Otherwise stripe-ruby 5.0 seems to works OK for most of my use cases.

### DIFF
--- a/lib/stripe_mock/api/errors.rb
+++ b/lib/stripe_mock/api/errors.rb
@@ -59,6 +59,12 @@ module StripeMock
 
       error_values.last.merge!(json_body: { error: json_hash }, http_body: { error: json_hash })
 
+      # In stripe-ruby > 5.0 Stripe::CardError.new third argument was moved
+      # to keyword argument `code`.
+      if Gem::Version.new(Stripe::VERSION) >= Gem::Version.new('5')
+        error_values.last.merge!(code: error_values.delete_at(2))
+      end
+
       error_values
     end
   end


thijsc:
Would be good to get this merged even though the test suite is very broken.

j15e:
Yeah the test suite is in a pretty bad shape, I didn't bother to add a test, I still feel bad about it ahah. This project is pretty useful but requires a lot of maintenance from non-paid Stripe contributors

JeremyKirkham:
Would be great to get this merged, it's effectively blocking me from testing anything other than the happy path for charges!

cody-elhard:
In my case I was able to fix a failing test by replacing

`StripeMock.prepare_card_error(:card_declined)`
with

`error = Stripe::CardError.new('Test Msg', 'param_name', code: 'bad_code', http_status: 444, http_body: 'body')`

`StripeMock.prepare_error(error)
`

I was able to rescue from `Stripe::CardError` which allowed my code to still work

j15e:
PS this should probably only be merged when we have a CI testing for `stripe-ruby > 5` but quite a few other specs also fail with 5 (mostly because `all` methods were renamed `list`).

alexmamonchik:
@j15e #658 has been merged. I don't think that we have to care about stripe version < 5. Maybe that's why this PR should be closed?

j15e:
Yep seems legit to close! Thanks for the update.

PR #645, opened 2019-08-27
Update travis to only run on supported versions of ruby

gravitystorm:
I'd like to help get the test suite back into shape. I think the first step is to drop the old versions of ruby that are no longer supported from the travis config, since it's unlikely that anyone is actually using them. 

This PR updates the travis config to use the latest releases of ruby 2.4, 2.5 and 2.6, and remove a stray gem install that dates back to the ruby 1.9 era.

This doesn't fix the test suite, but I think it's a good first step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,10 @@ group: deprecated-2017Q2
 sudo: required
 language: ruby
 rvm:
-  - 2.0.0
-  - 2.1.10
-  - 2.2.7
-  - 2.3.4
+  - 2.4.6
+  - 2.5.5
+  - 2.6.3
 before_install:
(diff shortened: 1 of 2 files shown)


tadejm:
What do you think also locking the stripe gem to "< 5"? That should pass all the specs on Travis 💚. That could be final `2.5.x` release, where as version 3.0 would then add support for stripe >= 5.
At the time upgrading the stripe gem to version >= 5 and using `stripe-ruby-mock` will 💔 in tests relying on Stripe mock.

alexmamonchik:
we a going to merge #647 and your changes are included. That's why I want to close your PR. Thank you

gravitystorm:
@alexmamonchik Thanks, that's fine by me. I'm looking forward to seeing #647 merged soon!

PR #648, opened 2019-09-17
start to implement checkout session API

fauxparse:
This is a very basic start at implementing the checkout session API, just because I needed it in a project I'm working on and I didn't have much time. In particular, a lot of the existing tests failed when I ran them on my machine, and I didn't have much luck digging into why (I see that the build is also failing on master).

Happy to collaborate with someone to take this further, but this is as good a starting point as I can get for now (hopefully better than nothing!)


chuckphillips:
@fauxparse Given that Checkout Sessions are generally used with Stripe Checkout and Stripe Webhooks, might you be able to support `checkout.session.completed` via `StripeMock.mock_webhook_event` in `spec/fixtures/stripe_webhooks/`?

fauxparse:
> @fauxparse Given that Checkout Sessions are generally used with Stripe Checkout and Stripe Webhooks, might you be able to support `checkout.session.completed` via `StripeMock.mock_webhook_event` in `spec/fixtures/stripe_webhooks/`?

I'll see if I get some time...

chuckphillips:
@fauxparse I just read that users can add their own fixture. I tested it and it works, so not critical to add.
https://github.com/rebelidealist/stripe-ruby-mock#customizing-webhooks

PR #653, opened 2019-10-10
add invoice_settings on mock_customer

lumir:
Right now I'm getting `undefined method invoice_settings for #<Stripe::Customer:0x00007fb1e13ff198>` and looks like this part is missing on customer data.

### DIFF
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -134,6 +134,11 @@ module StripeMock
         discount: nil,
         account_balance: 0,
         currency: currency,
+        invoice_settings: {
+          default_payment_method: nil,
+          custom_fields: nil,
+          footer: nil
+        },
         sources: {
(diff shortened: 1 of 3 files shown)


lumir:
fixing specs

lumir:
@alexmamonchik @gilbert

excid3:
I think you'll also want to modify `verify_card_payment` in the subscriptions request handler so it doesn't throw an error if there's a default payment method set on the customer.

```
# lib/stripe_mock/request_handlers/subscriptions.rb
      def verify_card_present(customer, plan, subscription, params={})
        return if customer[:default_source]
        return if customer[:invoice_settings][:default_payment_method]
````

lumir:
done @excid3 :)

excid3:
Thanks for merging @alexmamonchik 🤘

PR #664, opened 2019-11-10
Bugfix: pass through PaymentIntent amount to mocked Charge

typeoneerror:
If a successful charge is made via PaymentIntent, the amount of the Charge should generally match the PaymentIntent's amount. 

I'm not sure the PaymentIntent API is quite there yet, but this seemed like a useful temporary hack.


lib/stripe_mock/request_handlers/payment_intents.rb
```diff
@@ -172,7 +172,11 @@ def last_payment_error_generator(code: nil, message: nil, decline_code: nil)
 
       def succeeded_payment_intent(payment_intent)
         payment_intent[:status] = 'succeeded'
-        payment_intent[:charges][:data] << Data.mock_charge
+
+        payment_intent[:charges][:data] << Data.mock_charge(
```
alexmamonchik:
```suggestion
        payment_intent[:charges][:data] << Data.mock_charge(
          balance_transaction: btxn,
```

PR #687, opened 2020-01-09
Add PaymentIntent Webhooks

klaustopher:
This adds webhook payloads for `payment_intent.succeeded` and `payment_intent.payment_failed`.

It uses SEPA direct debit payment methods, I hope this does not break anybodys workflow


mnin:
Should we also add all `payment_intent` webhooks to be feature complete?

Screenshot 2020-01-16 at 12 02 19

PR #695, opened 2020-02-05
Add `has_more` attribute to all ListObject instances

gbp:
This method is called in Stripe when using the `auto_paging_each`
ListObject method. Without this being in the mock data this method
doesn't work.

Steps to reproduce:

```
$ rails console test
irb> StripeMock.start
irb> Stripe.api_key = 'sk_test_...'
irb> h = StripeMock.create_test_helper
irb> customer = Stripe::Customer.create(source: h.generate_card_token)
irb> customer.subscriptions.auto_paging_each.map(&:id)
```

### DIFF
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -142,12 +142,14 @@ module StripeMock
         sources: {
           object: "list",
           total_count: sources.size,
+          has_more: false,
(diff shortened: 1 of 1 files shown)


phedinkus:
+1 on needing this

phedinkus:
Thanks @alexmamonchik!

PR #698, opened 2020-02-06
Replaced succeeded payment intent charge mock with instance of Charge…

stefatkins:


tdeschamps:
Seems like it needs some refinements to make the tests ✅

lib/stripe_mock/request_handlers/payment_intents.rb
```diff
@@ -169,7 +169,9 @@ def succeeded_payment_intent(payment_intent)
         payment_intent[:status] = 'succeeded'
         btxn = new_balance_transaction('txn', { source: payment_intent[:id] })
 
-        payment_intent[:charges][:data] << Data.mock_charge(
+        payment_intent[:charges][:data] << Stripe::Charge.create(
```
tdeschamps: Not sure you should change `Data.mock_charge` to `Stripe::Charge.create`

stefatkins: This is to be able to use  Stripe::Charge.retrieve with id of the charge linked to the payment intent.

klaustopher: The only problem is that `Data.mock_charge` generates the Hash with the details, whereas `Stripe::Charge.create` gives you the `Charge` object. So, changing this to `Stripe::Charge.create(...).to_hash` should be the correct approach.

I agree with the general idea here 👍

PR #704, opened 2020-02-26
closes #702 - Allow the Plan.amount to be nil if the plan has tiers

### LINKED ISSUE #702: stripe-ruby-mock doesn't allow Plan.amount to be nil
The stripe api says Plan.amount should be `positive integer or zero` (https://stripe.com/docs/api/plans/object#plan_object-amount) - yet for Plans with pricing tiers that field is `null`.

stripe-ruby-mock currently has a number of checks to ensure that Plan.amount is a positive integer field, which breaks tests using Plans with pricing tiers.

pcraston:
Also add a subscription helper that calculates the amount based on the tiers

### DIFF
--- a/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
+++ b/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
@@ -111,7 +111,31 @@ module StripeMock
 
       def total_items_amount(items)
         total = 0
-        items.each { |i| total += (i[:quantity] || 1) * i[:plan][:amount] }
+        items.each do |item|
+          if not item[:plan][:amount].nil?
+            total += item[:quantity] * item[:plan][:amount]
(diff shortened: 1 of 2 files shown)


lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb
```diff
@@ -111,7 +111,31 @@ def verify_trial_end(trial_end)
 
       def total_items_amount(items)
         total = 0
-        items.each { |i| total += (i[:quantity] || 1) * i[:plan][:amount] }
+        items.each do |item|
+          if not item[:plan][:amount].nil?
```
alexmamonchik: we don't follow ruby style guide yet, but can you run rubocop locally exactly for your changes?
Also can it be modified with using `unless` and `next` commands to avoid constructions like `if not something.nil?`

PR #718, opened 2020-05-07
Subscription API update (2019-10-17): start has been replaced with start_date

lucianopanaro:
https://stripe.com/docs/upgrades#2019-10-17

> The start field on a subscription resource has been removed and is replaced by a start_date
> field which represents when the entire subscription started as opposed to when the current
> plan configuration started.


TastyPi:
This is fixed in #816

PR #727, opened 2020-06-11
Fix List initialize error with deleted records

jmulieri:
When deleting records, the record is marked internally as deleted but
other data fields are cleared, including `created`. The deleted record
is passed to StripeMock::Data::List.new from their respective #list
methods, e.g. Stripe::Customer.list. The sorting of these records in
initialize raised an ArgumentError (comparison of Integer with nil
failed).

This fix addresses the issue by omitting deleted records from the data
as that seems to be the more intuitive behavior. Minor refactoring was
also performed to normalize the sorting of data whether it is in Hash
or StripeObject form.

### DIFF
--- a/lib/stripe_mock/data/list.rb
+++ b/lib/stripe_mock/data/list.rb
@@ -9,12 +9,9 @@ module StripeMock
         @starting_after = options[:starting_after]
         @ending_before  = options[:ending_before]
         @active = options[:active]
-        if @data.first.is_a?(Hash) && @data.first[:created]
-          @data.sort_by! { |x| x[:created] }
-          @data.reverse!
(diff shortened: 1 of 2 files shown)


kylefox:
Thanks for this fix @jmulieri, it worked great for me 👍

Any chance this could get merged @alexmamonchik? 🙏 

My temporary workaround for retrieving deleted objects is to stub Stripe API methods:

```ruby
before do
  allow(Stripe::Customer).to(
    receive(:retrieve).once.with('cus_DELETED')
  ).and_return(
    Stripe::Customer.construct_from(id: 'cus_DELETED', deleted: true)
  )
end
```

PR #730, opened 2020-06-25
support on price api

hidenba:
The implementation maintains minimal compatibility with the planAPI


pbeckinger:
Thanks for the PR. I think https://github.com/stripe-ruby-mock/stripe-ruby-mock/blob/master/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb#L114 also needs to be updated since price object doesnt use `amount`.

jamesprior:
> Thanks for the PR. I think https://github.com/stripe-ruby-mock/stripe-ruby-mock/blob/master/lib/stripe_mock/request_handlers/helpers/subscription_helpers.rb#L114 also needs to be updated since price object doesnt use `amount`.

Would it be as straightforward as something like?
```ruby
items.each do |item| 
  quantity = item[:quantity] || 1
  amount = item[:price].present? ? item[:price][:unit_amount]: item[:plan][:amount]
  total += quantity * amount
end
```

I'd love to see this get merged in.

Asdolo:
So what's the state of this issue?

csalvato:
I just spent a few hours writing a patch for this, but then found that @hidenba has written it already!

Happy to see it here, and hopefully this can get merged in.

What will it take to get either this or #758 merged in?  @gilbert have you abandoned the project?  It's cool if you have, but maybe we need to make a community fork, if that's the case.

PR #731, opened 2020-06-26
Allow creating plan with nested product params for new product

cgunther:
Stripe allows you to either specify an existing product ID when creating a plan, or pass a hash of params to create a new service product alongside the new plan.

#595 (and #536 which it was based on) tried to support this, but an alternate PR was merged to support products.

Docs: https://stripe.com/docs/api/plans/create#create_plan-product


jonatasdaniel:
Any update on this? I'm having the same issue and it seems it's not fixed yet.

PR #732, opened 2020-06-29
Adds some basic call/response mocking of refunds for payment intents

### LINKED ISSUE #714: Unable to refund a payment intent
I am getting this error from this code

```
     Stripe::InvalidRequestError:
       No such charge:
```

```
let(:payment_intent) {  Stripe::PaymentIntent.create(amount:  100, currency: "jpy", confirm: true) }

Stripe::Refund.create({ payment_intent: payment_intent.id })

```

hammerdr:
* Does not link refunds through charges like I would expect because
  creation of payment intents do not seem to create and/or link charges

Fixes #714

### DIFF
--- a/lib/stripe_mock/request_handlers/refunds.rb
+++ b/lib/stripe_mock/request_handlers/refunds.rb
@@ -18,7 +18,18 @@ module StripeMock
           end
         end
 
-        charge = assert_existence :charge, params[:charge], charges[params[:charge]]
+        if params[:payment_intent]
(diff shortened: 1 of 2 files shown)


swrobel:
I would assume this fixes #714 ... If so, you might want to add that to the description so that gets auto-closed if this is merged.

JustinSeidl:
@TastyPi @swrobel can we get this PR merged? It looks good to me and I am experiencing the same issue as @hammerdr with payment intents and refunds.

PR #734, opened 2020-07-14
Add trial_from_plan to Subscription params whitelist

pn-y:


mnin:
LGTM

@pn-y can you just add also a spec for this case?

PR #743, opened 2020-09-02
Type is not a required field on products

louim:
According to the [API documentation](https://stripe.com/docs/api/products/create) the only required attribute for a Product is the `name`. Defaults will be merged in after in any case.


csalvato:
Perfect! Can you please upgrade the CHANGELOG with these changes and add a test to make sure that type is not required?

Also, a link to the portion of the API that confirms that type is not required would be much appreciated.

louim:
Fixed by https://github.com/stripe-ruby-mock/stripe-ruby-mock/commit/4e1759d6f12aa59ee05cd2295b9f03e0718acccd, where the check for type was removed.

PR #760, opened 2020-12-15
Include payment_method in PaymentIntent charges

tekin:
A Payment Intent captured with a payment method should reflect the payment method in any charges created against it to match the behaviour of real Stripe.


tekin:
Hello! is anyone able to take a look at this one (freshly-rebased with master) /cc @TastyPi

PR #762, opened 2021-01-04
Support Stripe Connect by adding stripe_account header namespace for customers

csalvato:
The Stripe library allows us to use pretty much every API call with an optional `stripe_account` key/value pair that is the Stripe user ID of the connected account.  For example:

```ruby
Stripe::Customer.update(connected_customer.id, { some: 'hash with customer data' }, stripe_account: 'some_stripe_user_id_string' )
```

The current implementation of this mocking library does not correctly mock this API. It will always use the default Stripe API key, when it should determine which API key to use based on the `stripe_account` key.

Any data relative to that account should be namespaced accordingly for future retrieval later when using the same `stripe_account` data. 

For example, if you are creating a customer using  `stripe_account: customer_1_stripe_id`, then this customer should not appear in a `Customer.list` call using `strip_account: customer_2_stripe_id`.

This PR accomplishes this by namespacing the mock requests and data based on the `stripe_account` key...and if it's not provided, we default back to the app's Stripe API key.


csalvato:
What do I need to do to get this merged?

csalvato:
I've been using these changes for several months on a private branch so think they are good to merge. Merging this in.

PR #767, opened 2021-02-21
Fixing all tests and more

lpsBetty:
I fixed a few errors that i've got when running my project's test suites, e.g. 

```
NoMethodError: undefined method detect for nil:NilClass
```

And also that the currencies does not match, because it is setting the default currency to the mocked customer (which should not be set).

All tests that I've fixed ran successfully on my local machine.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ end
 
 group :test do
   gem 'rake'
+  gem 'dotenv'
 end
 
 gemspec
--- a/lib/stripe_mock/data.rb
+++ b/lib/stripe_mock/data.rb
@@ -140,7 +140,7 @@ module StripeMock
 
     def self.mock_customer(sources, params)
       cus_id = params[:id] || "test_cus_default"
-      currency = params[:currency] || StripeMock.default_currency
+      currency = params[:currency]
       sources.each {|source| source[:customer] = cus_id}
       {
         email: 'stripe_mock@example.com',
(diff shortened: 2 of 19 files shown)


csalvato:
@lpsBetty are there any failing tests on your local?

lpsBetty:
@csalvato no i fixed them all

csalvato:
@lpsBetty Thanks for this work.

When I run specs locally, I get several failures:

```
Finished in 11.31 seconds (files took 0.8676 seconds to load)
1129 examples, 462 failures, 39 pending
```

We will need to set up some CI (maybe GitHub actions) to test various environments. Once we do that, we can have a red/green suite and development on this gem should be a lot faster.

What command are you running to run the specs? Also, what Ruby version?  If you have a fully passing suite, we can start there with support for the gem.

lpsBetty:
@csalvato ruby version 2.6.2 as defined in `.ruby-version` and executing with `bundle exec rspec`

Can you tell me one of your failures? Did you run it within my repository lpsBetty/stripe-ruby-mock ?

csalvato:
@lpsBetty Thanks for helping me clear this up. It was definitely because of running ruby 3.0 on my local environment. 

It looks like the server doesn't support ruby 3.0 just yet, and there will need to be some changes to make it work. I don't see a reason to prioritize that, though, unless someone raises an issue on it.

Ideally, we set up github actions/travis/circle to run tests on 2.6.2 and a few other ruby versions automatically. 🤔 

Not sure if that's something you're willing to take on, but i can merge this in, in the meantime and we can plan for another release.

I'll create another release candidate with these changes and push a new version of the gem.

csalvato:
The gem should now be updated and accessible with a `bundle update` 👍🏼

lpsBetty:
@csalvato very cool, thank you! yes i can help you with CI, when i find time ! :)

PR #773, opened 2021-04-21
Type is not required when cloning a PaymentMethod

ianyamey:
When cloning a PaymentMethod, the type parameter is [not required when `payment_method` is specified](https://stripe.com/docs/api/payment_methods/create#create_payment_method-type)

### DIFF
--- a/spec/shared_stripe_examples/payment_method_examples.rb
+++ b/spec/shared_stripe_examples/payment_method_examples.rb
@@ -128,6 +128,17 @@ shared_examples 'PaymentMethod API' do
         expect { payment_method }.to raise_error(Stripe::InvalidRequestError)
       end
     end
+
+    context 'when a payment_method is supplied' do
+      let(:type) { nil }
+      let(:original) { Stripe::PaymentMethod.create(type: 'card') }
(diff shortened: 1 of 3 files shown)


spec/shared_stripe_examples/payment_method_examples.rb
```diff
@@ -128,6 +128,17 @@
         expect { payment_method }.to raise_error(Stripe::InvalidRequestError)
       end
     end
+
+    context 'when a payment_method is supplied' do
+      let(:type) { nil }
+      let(:original) { Stripe::PaymentMethod.create(type: 'card') }
+      let(:payment_method) { Stripe::PaymentMethod.create(payment_method: original.id, type: type) }
```
MaicolBen, line 135: In Stripe, you can do `payment_method.card` and will work, but not here since we didn't pass the type 😞

MaicolBen (commented):
Good work!

spec/shared_stripe_examples/payment_method_examples.rb
ianyamey: Is there anything that you'd like me to change before it's ready to merge?

PR #786, opened 2021-07-26
Improve the `Checkout::Session` API

TastyPi:
Now it creates a `PaymentIntent` or `SetupIntent` when using the `payment` or `setup` mode respectively. Also moved the creation code into `RequestHandlers::Checkout::Session` for consistency.

This is technically a breaking change, however the changes make it work more like the actual API, so it should be an improvement.


lib/stripe_mock/request_handlers/checkout_session.rb
```diff
@@ -3,9 +3,88 @@ module RequestHandlers
...
+            object: "checkout.session",
+            allow_promotion_codes: nil,
+            amount_subtotal: amount,
+            amount_total: amount,
+            automatic_tax: {
+              enabled: false,
+              status: nil
+            },
+            billing_address_collection: nil,
+            cancel_url: "http://localhost:3000/connect/stold4/pay_later",
```
louim, line 63: This seems like a bit specific (probably from a real app running locally?). Maybe something like 
```suggestion
            cancel_url: "http://localhost:3000/checkout_sessions/cancel",
```

lib/stripe_mock/request_handlers/checkout_session.rb
```diff
@@ -3,9 +3,88 @@ module RequestHandlers
...
+            payment_intent: payment_intent,
+            payment_method_options: {},
+            payment_method_types: ["bacs_debit", "card"],
+            payment_status: payment_status,
+            setup_intent: setup_intent,
+            shipping: nil,
+            shipping_address_collection: nil,
+            submit_type: nil,
+            subscription: nil,
+            success_url: "http://localhost:3000/connect/stold4/pay_later_callback?session_id={CHECKOUT_SESSION_ID}&date=2021-08-02",
```
louim, line 82:
```suggestion
            success_url: "http://localhost:3000/checkout_sessions/success",
```

lib/stripe_mock/request_handlers/checkout_session.rb
```diff
@@ -3,9 +3,88 @@ module RequestHandlers
...
+            total_details: nil,
+            url: nil
```
louim, line 84: It would be nice to have at least a fake url, like 
```suggestion
            url: "https://checkout.stripe.com/pay/cs_test_abc123"
```

spec/shared_stripe_examples/checkout_session_examples.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    expect(session.setup_intent).to_not be_empty
+    setup_intent = Stripe::SetupIntent.retrieve(session.setup_intent)
+    expect(setup_intent.payment_method_types).to eq(["card"])
+  end
+
+  context "retrieve a checkout session" do
+    let(:checkout_session1) { stripe_helper.create_checkout_session }
+
+    it "ca be retrieved by id" do
```
louim, line 39:
```suggestion
    it "can be retrieved by id" do
```

louim (commented):
Thanks for the changes, this is something that would be very useful. I left some small comments.

lib/stripe_mock/request_handlers/checkout_session.rb
TastyPi: Oops, yeah these were from when I copy+pasted the JSON from testing the API. Changed.

louim (approved):
This looks good to me 🚀

PR #790, opened 2021-09-24
Fail with Stripe::InvalidRequestError instead of undefined method '[]' for nil:NilClass

MaicolBen:
With the new change in #762 of namespacing
stripe accounts, now the code fails differently
if the stripe account namespace is not set

### DIFF
--- a/lib/stripe_mock/request_handlers/customers.rb
+++ b/lib/stripe_mock/request_handlers/customers.rb
@@ -61,7 +61,7 @@ module StripeMock
       def update_customer(route, method_url, params, headers)
         stripe_account = headers && headers[:stripe_account] || Stripe.api_key
         route =~ method_url
-        cus = assert_existence :customer, $1, customers[stripe_account][$1]
+        cus = assert_existence :customer, $1, customers.dig(stripe_account, $1)
 
         # get existing and pending metadata
         metadata = cus.delete(:metadata) || {}
@@ -112,10 +112,10 @@ module StripeMock
       def delete_customer(route, method_url, params, headers)
         stripe_account = headers && headers[:stripe_account] || Stripe.api_key
         route =~ method_url
-        assert_existence :customer, $1, customers[stripe_account][$1]
+        assert_existence :customer, $1, customers.dig(stripe_account, $1)
 
         customers[stripe_account][$1] = {
-          id: customers[stripe_account][$1][:id],
+          id: customers.dig(stripe_account, $1)[:id],
           deleted: true
         }
       end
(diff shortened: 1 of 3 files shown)


lib/stripe_mock/request_handlers/customers.rb
```diff
@@ -61,7 +61,7 @@ def new_customer(route, method_url, params, headers)
       def update_customer(route, method_url, params, headers)
         stripe_account = headers && headers[:stripe_account] || Stripe.api_key
         route =~ method_url
-        cus = assert_existence :customer, $1, customers[stripe_account][$1]
```
MaicolBen, line 64: This failed `undefined method `[]' for nil:NilClass` instead of `Stripe::InvalidRequestError` when looking for a customer in a stripe account

lib/stripe_mock/request_handlers/tokens.rb
```diff
@@ -16,15 +16,13 @@ def create_token(route, method_url, params, headers)
 
         cus_id = params[:customer]
 
-        if cus_id && params[:source]
-          customer = assert_existence :customer, cus_id, customers[stripe_account][cus_id]
+        customer = assert_existence :customer, cus_id, customers.dig(stripe_account, cus_id) || customers[Stripe.api_key][cus_id]
```
MaicolBen, line 19: When [creating a token for cloning sources](https://stripe.com/docs/connect/cloning-customers-across-accounts), it's needed to create a token with a platform customer and a stripe account, so it isn't always true that the customer has to live in the same account.

lib/stripe_mock/request_handlers/tokens.rb
```diff
@@ -16,15 +16,13 @@ def create_token(route, method_url, params, headers)
...
 
+        if cus_id && params[:source]
           # params[:card] is an id; grab it from the db
           customer_card = get_card(customer, params[:source])
           assert_existence :card, params[:source], customer_card
         elsif params[:card].is_a?(String)
-          customer = assert_existence :customer, cus_id, customers[stripe_account][cus_id]
```
MaicolBen, line 26: refactoring this code to above since is repeated in each case

lib/stripe_mock/request_handlers/tokens.rb
```diff
@@ -34,16 +32,13 @@ def create_token(route, method_url, params, headers)
...
-
           # params[:bank_account] is an id; grab it from the db
           bank_account = verify_bank_account(customer, params[:bank_account])
           assert_existence :bank_account, params[:bank_account], bank_account
         elsif params[:bank_account]
           # params[:card] is a hash of cc info; "Sanitize" the card number
           bank_account = params[:bank_account]
         else
-          customer = assert_existence :customer, cus_id, customers[stripe_account][cus_id] || customers[Stripe.api_key][cus_id]
```
MaicolBen, line 46: The only case that handled correctly the source cloning, but if you want to clone a bank account or card, it won't work.

olimart:
Been awhile but still applicable unfortunately.
We only checked for the `tokens` file but this is a valid change indeed.

Issue #791, opened 2021-09-28
undefined method `require_params' when creating checkout sessions for payments and subscriptions


mpmenne:
Hey All,

I noticed that I was receiving the following error when I was creating a `Stripe::Checkout::Session` without using the default params on `StripeMock#create_checkout_session`
```
undefined method `require_params'
```

It looks like a two lines snuck into `master` without test coverage.  Here is one [spot](https://github.com/stripe-ruby-mock/stripe-ruby-mock/blob/a3bd40adbdfe01fb74af7489e7c269fed4246b1b/lib/stripe_mock/request_handlers/checkout_session.rb#L59) and here is [another spot](https://github.com/stripe-ruby-mock/stripe-ruby-mock/blob/a3bd40adbdfe01fb74af7489e7c269fed4246b1b/lib/stripe_mock/request_handlers/checkout_session.rb#L80) where the issue occurs.

This error is fixed by changing `require_params` to `require_param`.

I realize that this issue doesn't change the behavior of the library, but it may help the next person who sees this error. 

Happy to submit a PR to address if people are open to merging it.

PR #792, opened 2021-09-28
Fixes `undefined method 'require_params'` when creating checkout session

mpmenne:
This PR addresses an issue where `require_params` was being called instead of `require_param`.  

Previously, when the aforementioned lines were run, an `undefined method` error was thrown.


mpmenne:
Addresses #791

TastyPi (approved):

PR #797, opened 2021-10-28
Fix upcoming invoice failing when there was no customer

brunvez:
There's an issue where if we try to retrieve an invoice with just a subscription an error is raised saying that the customer needs to be present. So `Stripe::Invoice.upcoming(subscription: 'sub_123')` works with the Stripe API but not in the tests.

This PR fixes that issue by using the subscription's customer as a fallback.


vinc:
This fix an issue I'm also encountering 👍

PR #799, opened 2021-11-02
Allow events to be filtered by created

adamstegman:
See https://stripe.com/docs/api/events/list#list_events-created for API documentation.

Let me know if the tests seem like they might be flaky, I can add Timecop to mock the times to ensure they line up. I think they should be OK as-is though.

### DIFF
--- a/lib/stripe_mock/request_handlers/events.rb
+++ b/lib/stripe_mock/request_handlers/events.rb
@@ -4,7 +4,7 @@ module StripeMock
 
       def Events.included(klass)
         klass.add_handler 'get /v1/events/(.*)', :retrieve_event
-        klass.add_handler 'get /v1/events',      :list_events 
+        klass.add_handler 'get /v1/events',      :list_events
       end
 
       def retrieve_event(route, method_url, params, headers)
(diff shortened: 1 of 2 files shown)


lib/stripe_mock/request_handlers/events.rb
```diff
@@ -13,9 +13,36 @@ def retrieve_event(route, method_url, params, headers)
...
+          if params[:created][:lt]
+            event_list.select! { |event| event[:created] < params[:created][:lt].to_i }
+          end
+          if params[:created][:lte]
+            event_list.select! { |event| event[:created] <= params[:created][:lte].to_i }
+          end
+        else
+          event_list.select! { |event| event[:created] == params[:created].to_i }
+        end
+        event_list
```
TastyPi, line 43: You shouldn't use `select!` as this will modify the list and you haven't made a copy of it. Use `select` instead.

```suggestion
        if params[:created].is_a?(Hash)
          if params[:created][:gt]
            event_list.select { |event| event[:created] > params[:created][:gt].to_i }
          end
          if params[:created][:gte]
            event_list.select { |event| event[:created] >= params[:created][:gte].to_i }
          end
          if params[:created][:lt]
            event_list.select { |event| event[:created] < params[:created][:lt].to_i }
          end
          if params[:created][:lte]
            event_list.select { |event| event[:created] <= params[:created][:lte].to_i }
          end
        else
          event_list.select { |event| event[:created] == params[:created].to_i }
        end
```

TastyPi (changes requested):

lib/stripe_mock/request_handlers/events.rb
adamstegman: The list we're modifying came from `events.values`, which itself is a new copy—there's no need to `dup` it again.

adamstegman: I went ahead and changed it though, so any future refactoring doesn't have to worry about it.

lib/stripe_mock/request_handlers/events.rb
```diff
@@ -13,9 +13,36 @@ def retrieve_event(route, method_url, params, headers)
...
+          if params[:created][:lt]
+            event_list = event_list.select { |event| event[:created] < params[:created][:lt].to_i }
+          end
+          if params[:created][:lte]
+            event_list = event_list.select { |event| event[:created] <= params[:created][:lte].to_i }
+          end
+        else
+          event_list = event_list.select { |event| event[:created] == params[:created].to_i }
+        end
+        event_list
```
TastyPi, line 43: nit: you could remove the assignments and the last `event_list` so the if statement returns the value, like I had in my original suggestion. It's a bit less repetitive, but if you think this is more readable I don't care strongly.

adamstegman: These `if`s don't return, they continue changing the result. You can apply multiple filters and all are considered when filtering results, e.g. `created: {gt: time1, lt: time2}` will only give you events between those two timestamps.

TastyPi: Ah you're right, I misread the code.

TastyPi (approved):

PR #800, opened 2021-11-19
Fixes billing_cycle_anchor attribute on Subscription to accept `now` as value

fabianoarruda:
Hey guys, this is my first time contributing to this awesome library. It has been very useful to me, so I wanted to start giving back as much as possible.

This PR will fix the attribute `billing_cycle_anchor`. Usually you pass a timestamp as value, but you should also be able to pass the string `now` and the billing cycle will be reset immediately. ( [Stripe documentation reference ](https://stripe.com/docs/billing/subscriptions/billing-cycle)) 
When you send `now` in this attribute, Stripe API will respond with this attribute converted to current timestamp.

It will also add missing attributes to subscription:
- `proration_behavior` 
- `backdate_start_date`


TastyPi:
Please add a test

fabianoarruda:
@TastyPi Test added!

TastyPi (approved):

PR #801, opened 2021-11-21
Set the status to "incomplete" when Stripe::Subscription's payment_behavior is "default_incomplete".

NerdyBoyCool:
I tried to create a subscription which is payment_behavoir is deufalt_impcomplete, but ``StripeMock::RequestHandlers::Subscriptions::verify_card_present`` block my request.

I added "incomplete" status to subscirption when "default_incomplete" is passed.

@see https://stripe.com/docs/billing/subscriptions/build-subscription?ui=elements#create-subscription
You will find instructions on how to create a subscription without specifying a payment method.
In this doc, the payment and billing is finalized at the FrontEnd, and the "incomplete" subscription is created server-side first.

@see https://stripe.com/docs/api/subscriptions/create#create_subscription-payment_behavior

> Use default_incomplete to create Subscriptions with status=incomplete when the first invoice requires payment, otherwise start as active.

### DIFF
--- a/lib/stripe_mock/request_handlers/subscriptions.rb
+++ b/lib/stripe_mock/request_handlers/subscriptions.rb
@@ -132,6 +132,10 @@ module StripeMock
           subscription[:status] = 'trialing'
         end
 
+        if params[:payment_behavior] == 'default_incomplete'
+          subscription[:status] = 'incomplete'
+        end
+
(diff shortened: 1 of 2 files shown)


NerdyBoyCool:
@TastyPi Thank you so much!

PR #820, opened 2022-02-20
Allow expanding default_source on customer object

fabianoarruda:
This PR adds support for expanding the default source on customer object. Useful for when you want to get a customer object with the default Credit Card, without the need of a second call on Stripe API to fetch Card details.


fabianoarruda:
I have fixed the conflicts.

PR #823, opened 2022-03-22
Allow user to pass in a subscription ID on create

rbclark:
A few lines down `params[:id]` is used, however since it is not in the allowed params list it is not possible to take the code path to create a subscription with an existing param. This allows the ID param through.

https://github.com/stripe-ruby-mock/stripe-ruby-mock/blob/3a14a8bd7e9e2e259fcc64987c852b4d8146d54c/lib/stripe_mock/request_handlers/subscriptions.rb#L106


barsbek:
Hey guys. Can we merge this? Would be really helpful for our integration tests

PR #824, opened 2022-03-23
Add fixture for invoice.paid webhook

rbclark:
This is a webhook that is very similar to invoice.payment_succeeded however it occurs whenever an invoice payment attempt succeeds or an invoice is marked as paid out-of-band. invoice.payment_succeeded only fires when an invoice payment attempt succeeds.


rbclark:
Superseded by https://github.com/stripe-ruby-mock/stripe-ruby-mock/pull/825

PR #830, opened 2022-05-11
Implement search API

adamstegman:
Stripe has added a [search API](https://stripe.com/docs/search) to the latest API version, `2020-08-27`. Supporting this in all its permutations would be very difficult, essentially partially implementing a full search engine. For now, supporting exact-match searches on each of the supported fields is relatively easy.


prayeshshah:
Hey @adamstegman , I also require this functionality. Let me know if there is anything I can do to move this PR forward.

ryenski:
This [works well for me](https://github.com/MemberDrive/stripe-ruby-mock) and IMO should be an easy merge.

PR #834, opened 2022-06-04
Allow expanding default_source on customer object

fabianoarruda:
This PR adds support for expanding the default source on customer object. Useful for when you want to get a customer object with the default Credit Card, without the need of a second call on Stripe API to fetch Card details.

In addition to that, I added some missing attributes for `mock_charge` and reorganized to be in alphabetical order.


fabianoarruda:
All tests now passing 👍

fabianoarruda:
@alexmamonchik Yes, I think so. The library already supports expanding `default_source` when retrieving an existing customer. This PR allows to expand it when creating the customer also. I edited the description of the PR to clarify that.

We already have a test for expanding `default_source` on `retrieve`. I'll add another test to assert that when passing the expand parameter with the `create` method, the customer object returned will contain the `default_source` expanded.

fabianoarruda:
@alexmamonchik I have updated the branch ahd added a new test as requested.

PR #842, opened 2022-09-07
Add the pay_immediately param to create subscription allowed params

trevorj06:
Stripe supports a custom parameter called "pay_immediately" that is only available by request to Stripe. Setting this flag to false will cause a subscription invoice to not be paid for one hour. This allows an external service (such as one for taxes) to make updates to the invoice before payment is made. This parameter needs to be added the allowed parameters list so that anyone using this feature will still have passing specs that rely on stripe ruby mock.


alexmamonchik:
looks like it's not a public method https://github.com/stripe/stripe-mock/issues/220 if you still need it, please, update PR and add some tests

PR #858, opened 2023-02-03
Allow currency param in create subscription

learod:


gregmolnar:
@learod can you add a changelog entry for this please?

PR #860, opened 2023-02-28
Adds error when attempting to refund a charge that has already been refunded / refunding more than the charge amount

jakebills1:
Throws InvalidRequestError if the sum of the refund amount plus what is already refunded on charge is greater than charge amount


alexmamonchik:
@jakebills1 please, reopen your PR and let's run tests here. It had failed tests after merge it to master

jakebills1:
@alexmamonchik Reopened it here: https://github.com/stripe-ruby-mock/stripe-ruby-mock/pull/889

PR #861, opened 2023-03-11
Fix test suite

smtlaissezfaire:
This fixes the test suite (ran locally under ruby 3.1.2).

I was still occasionally seeing DRB-style "is recycled object" errors; didn't look further into them.

Is the suite set up on some sort of CI?

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,2 @@
---colour --format documentation
+--colour
+--format progress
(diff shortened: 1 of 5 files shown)


gregmolnar:
Thanks for working on this! There is no CI yet, I started to look into options to set up one. If we want GitHub actions, the usage would be counted towards the admin of the organization and I don't think they would want that. I checked CircleCI and that would be free, but their GitHub integration asks for permission on all repositories I have access to. I would prefer something where I could add it to a single repository only. Do you know any such service?

smtlaissezfaire:
I think it should be OK to use github actions?  Isn't the org "stripe-ruby-mock" ?

PR #868, opened 2023-03-18
setup github actions for specs

gregmolnar:


gregmolnar:
Scratch this, because it will count the minutes towards the organisation owner. I will setup circle ci instead.

PR #883, opened 2023-09-29
Allow stripe-ruby 9

joemsak:
Our test suite seems okay with just bumping the gemspec. YMMV


alexmamonchik:
we are going to release v4 and this changes will be covered by #893

PR #889, opened 2023-10-25
Adds error when attempting to refund a charge that has already been refunded / refunding more than the charge amount

jakebills1:
Reverts stripe-ruby-mock/stripe-ruby-mock#888
Readds changes to throw error when attempting to refund a charge for more than has been charged already or more than the original charge amount.


alexmamonchik:
@jakebills1 can you fix tests for this PR?

PR #903, opened 2024-04-10
Add `require` option to require additional file

stevenharman:
When booting the stripe-mock-server it can be useful to have Ruby `require` in additional files if you need to patch `StripeMock` in some way. This is mostly useful for differences when you're running on an older version of the Stripe API and need to tweak `StripeMock`'s behavior until your Stripe API version is upgraded to be compatible.

When using `StripeMock` directly within a Rails app, for example, this is unneeded as you can do the same thing via a Rails initializer file. But if running the `StripeMock` server standalone - for use with front end tests, for example, you might pass this new flag like this:

```console
--require config/initializers/stripe_ruby_mock
```

### DIFF
--- a/bin/stripe-mock-server
+++ b/bin/stripe-mock-server
@@ -11,6 +11,7 @@ opts = Trollop::options do
   opt :server, "Server to use", :type => :string, :default => 'thin'
   opt :debug, "Request and response output", :default => true
   opt :pid_path, "Location to put server pid file", :type => :string, :default => './stripe-mock-server.pid'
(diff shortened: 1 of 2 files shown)


stevenharman:
@alexmamonchik I've rebased this atop the latest `master`. Any chance of getting this merged?

stevenharman:
Thank you, @alexmamonchik. Any chance of cutting a 4.1 release with all of these improvements?

alexmamonchik:
@stevenharman sure. I would like to add few more changes and then release.

PR #905, opened 2024-04-19
Allow Stripe SDK v11

stevenharman:
Which launched a couple weeks ago, and should be backward compatible with 10, etc…

[Strike SDK Changelog](https://github.com/stripe/stripe-ruby/blob/master/CHANGELOG.md#1100---2024-04-10)


chrisjeon:
Can we get this in?

PR #919, opened 2024-08-13
Add capabilities to mock account data

cazwazacz:
### Why
Stripe returns a `capabilities` property on the `Account`, [shown here](https://docs.stripe.com/api/accounts/object#account_object-capabilities).

I wasn't sure if I needed to add a test, from a quick look I couldn't see tests for other properties on the mock data. More than happy to add a test if required!


alexmamonchik:
hi @cazwazacz Thank you for your help. Please check discussion #915 and tell us what do you think

PR #925, opened 2025-02-05
Fix github actions and add missing product attributes

bettysteger:


bettysteger:
The failing test on ruby 2.7 is not coming from my code changes...

PR #931, opened 2025-02-24
Support Stripe version 13

gabrieltaylor:
# Description
Adds support for Stripe versions >= 13.0.0 while maintaining support for version 12.

In Stripe version 13 `Stripe::StripeClient` was replaced by `Stripe::APIRequestor`.  This change adds support for version 13 and verifies that tests pass in both version 12 and 13 using the [Appraisal Gem](https://github.com/thoughtbot/appraisal).

To run the specs you can use the commands

```
bundle exec appraisal stripe-12 rspec spec 
```
```
bundle exec appraisal stripe-13 rspec spec 
```

Version logic is largely extracted to the new module `StripeMock::Compat` with the exception of instances where the logic is based on the number of arguments to a method.

This change also addresses a number of warnings in the specs related to the use of the `raise_error` matcher without specifying the class of the error raised.

# TODO
- [ ] Update GitHub actions to run specs against multiple versions of the Stripe gem.


lukeasrodgers:
@gabrieltaylor thank you for this work!

One interesting thing to note: at least on my system, specs run about 10x slower with this PR. At first this was pretty confusing and surprising. I believe the cause is actually the Stripe gem's use of `Stripe::APIRequestor::SystemProfiler.uname` which shells out and is quite expensive. 

Monkeypatching that in `spec/spec_helper.rb` fixes the performance issue. 

```ruby
Stripe::APIRequestor::SystemProfiler.class_eval do
  def self.uname
    "stripe-mock"
  end
end
```

Obviously that is an ugly hack, but imo we should do *something* like this, since the specs taking > 1minute to run significantly impedes development. Alternatively it may be possible to reuse instances of `APIRequestor` which would mitigate the need for a monkeypatch.

gabrieltaylor:
@lukeasrodgers great catch! Thank you for bringing that to my attention. I'll look into some of the solutions you suggest but if you feel inspired to make the change yourself I would gladly welcome your contribution.

lukeasrodgers:
@gabrieltaylor seems like this would work and is less hacky:

```ruby
# add to StripeMock::Compat
def self.client_instance
  @client ||= Stripe::APIRequestor.new
end

# add to StripMock::Instance
client = Compat.client_instance
```

feel free to add to the PR if it looks good to you

gabrieltaylor:
Love it! Will do.

alexmamonchik:
looks like not so small changes. Should we release it with version 5.0.0 of StripeRubyMock?

gabrieltaylor:
> looks like not so small changes. Should we release it with version 5.0.0 of StripeRubyMock?

That seems reasonable to me. There may be inadvertent breaking changes.

gabrieltaylor:
@lukeasrodgers @alexmamonchik I've updated the GH action to run the specs against both Stripe 12 and Stripe 13. Please let me know if there are any other changes you'd like to see added.

alexmamonchik:
> @lukeasrodgers @alexmamonchik I've updated the GH action to run the specs against both Stripe 12 and Stripe 13. Please let me know if there are any other changes you'd like to see added.

@gabrieltaylor you did the great job! Than you. Can you check failed actions? I think we can release this PR as a 5.0.0 ASAP

gabrieltaylor:
@alexmamonchik I believe I have corrected the CI issue. My branch was behind main. Can you please trigger the workflow run?

PR #940, opened 2025-06-26
Add discounts to create_subscription allowed_params

imella:
Added additional `allowed_params` to `create_subscription` to support `discounts`.


alexmamonchik:
should it also be in mock object? https://github.com/stripe-ruby-mock/stripe-ruby-mock/blob/master/lib/stripe_mock/data.rb#L389

imella:
@alexmamonchik Yeah, it would be useful to have it there. What do you guys usually do in this case? The `discounts` field is expandable, so do you put an array of ids or the whole object as if it was expanded?

PR #961, opened 2026-05-18
Add ruby 3.4 to the test matrix

gravitystorm:
I've run into some warnings running this gem on ruby 3.4. But before I try fixing any of them, I think it's best to start testing ruby 3.4 in CI.


gravitystorm:
If there's anything I can do to move this along, please do let me know! I'm keen to help. 

I think this PR is already pretty low-risk and a minimal change, but if I'm overlooking anything I'm all ears.

gravitystorm:
Thanks @alexmamonchik !

PR #962, opened 2026-06-17
Remove unused require 'ostruct'

gravitystorm:
The require was added in c34c4e8 (2013) but was apparently never used.

Removing it silences the Ruby 3.4 bundled-gem deprecation warning ("ostruct was loaded from the standard library, but will no longer be part of the default gems") without needing to add ostruct as a dependency.


alexmamonchik:
There are a lot of outdated things here. Thank you for the cleanup!

PR #964, opened 2026-06-18
Fix duplicate "status" key in verified bank account data

gravitystorm:
Elsewhere in the code the symbol key `:status` is used, and this mismatch is flagged up as a warning when using ruby 4.0


gravitystorm:
With #966 merged, I've rebased this PR and the CI is now green.

PR #965, opened 2026-06-18
Add ruby 4.0 to CI matrix

gravitystorm:
Everything works fine on ruby 4, the only thing that needed fixing was the logger dependency for the stripe gem. 

We can remove this line from the gemspec when we remove support for `stripe < 19` - which might be a while! :smile:


gravitystorm:
With #966 merged, I've rebased this PR and the CI is now green.

PR #967, opened 2026-06-20
MultiJson warning

espen:
The MultiJson constant is deprecated and will be removed in v2.0. Use MultiJSON instead. MultiJSON.load is deprecated and will be removed in v2.0. Use MultiJSON.parse instead.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,9 @@ source 'https://rubygems.org'
 group :test do
   gem 'rake'
   gem 'dotenv'
+  # stripe requires logger but does not declare it; logger is no longer a
+  # default gem as of Ruby 4.0
+  gem 'logger'
 end
 
 gemspec
--- a/gemfiles/stripe_12.gemfile
+++ b/gemfiles/stripe_12.gemfile
@@ -8,6 +8,7 @@ gem "stripe", "12.6.0"
 group :test do
   gem "rake"
   gem "dotenv"
+  gem "logger"
 end
 
 gemspec path: "../"
--- a/gemfiles/stripe_12.gemfile.lock
+++ b/gemfiles/stripe_12.gemfile.lock
@@ -16,32 +16,33 @@ GEM
       thor (>= 0.14.0)
     daemons (1.4.1)
     dante (0.2.0)
-    diff-lcs (1.6.1)
-    dotenv (2.8.1)
-    drb (2.2.1)
+    diff-lcs (1.6.2)
+    dotenv (3.2.0)
+    drb (2.2.3)
     eventmachine (1.2.7)
-    multi_json (1.15.0)
-    rack (2.2.13)
-    rake (13.2.1)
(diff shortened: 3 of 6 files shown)


gravitystorm:
All the tests failed, because the older versions of the multi_json gem used in the appraisals (1.15.0 and 1.19.1) don't have the new module name. The warnings, and the new module name, were both introduced in multi_json 1.21.0.

To make this PR work, you would also need to change the version constraint in the gemfile:

```diff
-  gem.add_dependency 'multi_json', '~> 1.0'
+  gem.add_dependency 'multi_json', '~> 1.21'
```

... and then regenerate all the appraisals.

I'm not sure if it's a good idea forcing everyone to use the latest version of multi_json (for example, I have an app which uses multi_json in various gems, and we have considered pinning to 1.20). I'm also not even sure if the complexity of using multi_json instead of stdlib json is still worthwhile.

espen:
Agree, given this gem is for testing I do not see the point of using MultiJson over stdlib.

I regenerated appraisals. I also added logger gem to make these tests run on Ruby 4

gravitystorm:
The failing tests are because the Appraisals configuration is missing many of the stripe versions - I've opened a PR to fix this at #968 .

gravitystorm:
I thought I should mention here that #968 is now merged, so a rebase (and regenerating the appraisals) should lead to a green CI now. If it doesn't, I'm happy to help debug further. The logger dependency was also merged in #965 so doesn't need to be included any more.

(I don't know whether merge commits in PRs is OK in this repo, so I would suggest choosing the side of caution and do a `git rebase master`, so that this PR only contains only the multi_json changes and no merge commits).

I'm still not sure whether it's better to restrict the multi_json minimum version, or drop the multi_json dependency entirely. I lean slightly towards the latter, but it's not my decision. Perhaps we could do the following?

* Update this PR so that CI is green
* Open an alternative PR, with multi_json dropped completely
* Ask @alexmamonchik which he prefers

That way it's as straightforward as possible for him to deal with - he would only need to merge one or the other PR.

alexmamonchik:
> I'm still not sure whether it's better to restrict the multi_json minimum version, or drop the multi_json dependency entirely. I lean slightly towards the latter, but it's not my decision. Perhaps we could do the following?
> 
> * Update this PR so that CI is green
> * Open an alternative PR, with multi_json dropped completely
> * Ask @alexmamonchik which he prefers
> 
> That way it's as straightforward as possible for him to deal with - he would only need to merge one or the other PR.

I prefer to remove old dependencies. This gem is already has a lot of outdated staff. Let's move on and support only newest things. It's impossible to just replace everything in one PR, of course. Thank you all for cleaning this. From my side, I'll merge and release new versions if you ping me.

gravitystorm:
We went for the alternative approach (dropping MultiJson completely) in #969, so this PR can be closed.

espen:
A better approach, thanks.

PR #968, opened 2026-07-03
Add missing appraisals configuration

gravitystorm:
PR #958 added the Gemfiles for Stripe v6-11, but didn't add the Appraisals configuration. This means that these Gemfiles won't be updated when running `bundle exec appraisal update`

I spotted this while trying to help get CI to go green in #967


gravitystorm:
Thanks @alexmamonchik !

